Issue #2, opened 2017-06-13
Make the effect type generic


gvolpe:
Right now the implementation is tied to cats.effect.IO. Make all the public methods generic using cats.effect.Effect[F[_]].

PR #5, opened 2017-06-13
#2 Effect type is now generic. 

gvolpe:
Still relying on cats.effect.IO for internal use.


gvolpe:
Still need to update the documentation (README file)

PR #35, opened 2017-07-07
Bind exchange

### LINKED ISSUE #31: Implement exchangeBind(String destination, String source, String routingKey)
Bare in mind that fs2-rabbit is just a wrapper for the AMQP Java client. So, whenever adding a new method that is supported by the native client, the procedure will be more or less the same:

1. Check the signatures in the java doc of the official java client:
https://www.rabbitmq.com/releases/rabbitmq-java-client/current-javadoc/com/rabbitmq/client/Channel.html
2. For this case, the return type is AMQP.Exchange.BindOk so in fs2-rabbit it will be Stream[F, Exchange.BindOk]
3. Add the new method at the end of the Fs2Rabbit file.
4. Add scala doc comments to the new method. Take a look at the existent methods but in general it could be more or less similar to the Java client docs. This is the one for this method: "Bind an exchange to an exchange, with no extra arguments."
5. Add a unit test in the file Fs2RabbitSpec: I can help you out with this part since it's not as trivial as adding the new method and sometimes it might be okay to just invoke the method on an existent test instead of creating a new one.

HINT: Look at the existent method [bindQueue](https://github.com/gvolpe/fs2-rabbit/blob/master/core/src/main/scala/com/github/gvolpe/fs2rabbit/Fs2Rabbit.scala#L242), this one will be very similar ;)

Thanks for contributing!

szymonm:
Resolves #31 

bindExchange test is ignored, because qpid does not support bindExchange.

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/EffectUnsafeSyncRunner.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/EffectUnsafeSyncRunner.scala
@@ -22,7 +22,7 @@ import scala.language.higherKinds
   * program.run.unsafeRunSync
   * }}}
   *
-  * Having this generic type allows the fs2-rabbit library to to this for any effect:
+  * Having this generic type allows the fs2-rabbit library to do this for any effect:
   *
   * {{{
   * import cats.effect.Effect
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/github/gvolpe/fs2rabbit/EffectUnsafeSyncRunner.scala
```diff
@@ -22,7 +22,7 @@ import scala.language.higherKinds
   * program.run.unsafeRunSync
   * }}}
   *
-  * Having this generic type allows the fs2-rabbit library to to this for any effect:
+  * Having this generic type allows the fs2-rabbit library to do this for any effect:
```
gvolpe: Thanks for fixing the typo :)

gvolpe:
Great! I'll merge it and I'll look into the qpid broker issue to see if can pick your unit test as well. Thanks!

PR #52, opened 2018-03-07
Add implicit execution context to Fs2Rabbit.apply

felixmulder:
This PR adds the ability to specify the internal queue's execution context.

I've had a look around the code and there are a couple of places where `unsafeRunSync` is being called. Maybe we can get around it somehow, but not sure as of now :)

I changed the signature to return an `F[Fs2Rabbit[F]]` similar to what's being done in fs2 for mutable queues and the like. I think it's a good idea to be pure all the way. See the examples for the code change needed, it's not so bad :)


gvolpe:
Looks great, thanks a lot @felixmulder !

~~EDIT: There's something wrong with the `StreamLoop` test now. The test coverage has dropped from 98% to 75%. I'll give it a try now but if can't figure it out quickly it might have to wait until the weekend.~~

EDIT 2: Some weird misleading report of `codecov`. I ran it locally and it's all good :+1:

PR #53, opened 2018-03-08
Harmonize execution contexts

felixmulder:
Originally there were two internal execution contexts, this PR
makes sure a single configurable EC is running fs2-rabbit.

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AmqpClientStream.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AmqpClientStream.scala
@@ -32,12 +32,11 @@ import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 
 class AmqpClientStream[F[_]](internalQ: mutable.Queue[IO, Either[Throwable, AmqpEnvelope]])(implicit F: Effect[F],
-                                                                                            SE: StreamEval[F])
+                                                                                            SE: StreamEval[F],
+                                                                                            EC: ExecutionContext)
     extends AMQPClient[Stream[F, ?]] {
 
-  private def defaultConsumer(channel: Channel): Consumer = {
-    implicit val queueEC: ExecutionContext = ExecutionContext.fromExecutor(Executors.newCachedThreadPool())
-
+  private def defaultConsumer(channel: Channel): Consumer =
     new DefaultConsumer(channel) {
 
       override def handleCancel(consumerTag: String): Unit =
@@ -52,9 +51,7 @@ class AmqpClientStream[F[_]](internalQ: mutable.Queue[IO, Either[Throwable, Amqp
         val props = AmqpProperties.from(properties)
         internalQ.enqueue1(Right(AmqpEnvelope(DeliveryTag(tag), msg, props))).unsafeRunSync()
       }
-
     }
-  }
 
   override def basicAck(channel: Channel, tag: DeliveryTag, multiple: Boolean): Stream[F, Unit] = SE.evalF {
     channel.basicAck(tag.value, multiple)
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/Fs2Rabbit.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/Fs2Rabbit.scala
@@ -38,7 +38,7 @@ object Fs2Rabbit {
       amqpClient <- F.delay(new AmqpClientStream[F](internalQ))
       config     <- F.delay(new Fs2RabbitConfigManager[F].config)
       connStream <- F.delay(new ConnectionStream[F](config))
-      fs2Rabbit  <- F.delay(new Fs2Rabbit[F](config, connStream, internalQ)(F, amqpClient))
+      fs2Rabbit  <- F.delay(new Fs2Rabbit[F](config, connStream, internalQ)(F, amqpClient, ec))
     } yield fs2Rabbit
 }
 // $COVERAGE-ON$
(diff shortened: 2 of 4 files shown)


felixmulder:
Codacy is complaining over 8 params in a single method. I agree with the sentiment that methods shouldn't have too many params ^^

But, I don't want to change the API.

gvolpe:
> Codacy is complaining over 8 params in a single method. I agree with the sentiment that methods shouldn't have too many params ^^
But, I don't want to change the API - and the added param is implicit.

It's fine. That's something that needs to be reworked. @insdami proposed to introduce `newtypes` for better encoding of the algebra but still nobody has done it. I might tackle it in my spare time.

core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/AMQPClient.scala
```diff
@@ -21,12 +21,14 @@ import com.github.gvolpe.fs2rabbit.model.ExchangeType.ExchangeType
 import com.github.gvolpe.fs2rabbit.model._
 import com.rabbitmq.client.Channel
 
+import scala.concurrent.ExecutionContext
+
```
gvolpe: The `algebra` and likely the `programs` should not be aware of an `ExecutionContext`. The only place where I'd put an EC would be in the in the `interpreters`.

core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/AckerConsumer.scala
```diff
@@ -19,17 +19,20 @@ package com.github.gvolpe.fs2rabbit.algebra
 import com.github.gvolpe.fs2rabbit.model.{AckResult, AmqpEnvelope, BasicQos, QueueName}
 import com.rabbitmq.client.Channel
 
+import scala.concurrent.ExecutionContext
```
gvolpe: Same here

core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/AMQPClient.scala
```diff
@@ -21,12 +21,14 @@ import com.github.gvolpe.fs2rabbit.model.ExchangeType.ExchangeType
...
 // format: off
 trait AMQPClient[F[_]] extends Binding[F] with Declaration[F] with Deletion[F] {
   def basicAck(channel: Channel, tag: DeliveryTag, multiple: Boolean): F[Unit]
   def basicNack(channel: Channel, tag: DeliveryTag, multiple: Boolean, requeue: Boolean): F[Unit]
   def basicQos(channel: Channel, basicQos: BasicQos): F[Unit]
-  def basicConsume(channel: Channel, queueName: QueueName, autoAck: Boolean, consumerTag: String, noLocal: Boolean, exclusive: Boolean, args: Map[String, AnyRef]): F[String]
+  def basicConsume(channel: Channel, queueName: QueueName, autoAck: Boolean, consumerTag: String, noLocal: Boolean, exclusive: Boolean, args: Map[String, AnyRef])(implicit EC: ExecutionContext): F[String]
```
gvolpe: Same here

core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Consuming.scala
```diff
@@ -19,16 +19,20 @@ package com.github.gvolpe.fs2rabbit.algebra
 import com.github.gvolpe.fs2rabbit.model.{AckResult, AmqpEnvelope, BasicQos, ConsumerArgs, QueueName}
 import com.rabbitmq.client.Channel
 
+import scala.concurrent.ExecutionContext
```
gvolpe: same here

core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AmqpClientStream.scala
```diff
@@ -35,9 +35,7 @@ class AmqpClientStream[F[_]](internalQ: mutable.Queue[IO, Either[Throwable, Amqp
                                                                                             SE: StreamEval[F])
     extends AMQPClient[Stream[F, ?]] {
 
-  private def defaultConsumer(channel: Channel): Consumer = {
-    implicit val queueEC: ExecutionContext = ExecutionContext.fromExecutor(Executors.newCachedThreadPool())
-
+  private def defaultConsumer(channel: Channel)(implicit EC: ExecutionContext): Consumer =
```
gvolpe: Probably it'd be better to add it as a class parameter (of the interpreter) if it can't be done at the method level without changing the signature of the algebras.

felixmulder:
@gvolpe - thanks for the feedback, I updated it in accordance with your recommendations :+1:

gvolpe:
Thanks @felixmulder , only asked you for one small change (revert changes actually). The rest looks good to me, will merge soon.

core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/AMQPClient.scala
```diff
@@ -48,4 +48,4 @@ trait Declaration[F[_]] {
 trait Deletion[F[_]] {
   def deleteQueue(channel: Channel, queueName: QueueName, ifUnused: Boolean = true, ifEmpty: Boolean = true): F[Unit]
   def deleteQueueNoWait(channel: Channel, queueName: QueueName, ifUnused: Boolean = true, ifEmpty: Boolean = true): F[Unit]
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
felixmulder: ah, missed that - remedied just now :)

felixmulder:
Thanks, I missed reverting that - should be good to go now :)

PR #54, opened 2018-03-08
Replace typesafe config library and unwrap the config from effect

insdami:
This is to replace config's library and unwrap it from effect, suggested by @felixmulder. Have a look and let me know :D


core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/Fs2Rabbit.scala
```diff
@@ -36,14 +34,14 @@ object Fs2Rabbit {
     for {
       internalQ  <- F.liftIO(fs2.async.boundedQueue[IO, Either[Throwable, AmqpEnvelope]](500))
       amqpClient <- F.delay(new AmqpClientStream[F](internalQ))
-      config     <- F.delay(new Fs2RabbitConfigManager[F].config)
+      config     <- Fs2RabbitConfig.load[F]
       connStream <- F.delay(new ConnectionStream[F](config))
       fs2Rabbit  <- F.delay(new Fs2Rabbit[F](config, connStream, internalQ)(F, amqpClient, ec))
     } yield fs2Rabbit
 }
```
felixmulder: Would you mind adding an apply method that takes the config as a parameter as well? :tada:

core/src/main/scala/com/github/gvolpe/fs2rabbit/program/AckerConsumerProgram.scala
```diff
@@ -44,7 +44,7 @@ class AckerConsumerProgram[F[_]](internalQ: mutable.Queue[IO, Either[Throwable,
     _.flatMap {
       case Ack(tag) => AMQP.basicAck(channel, tag, multiple = false)
       case NAck(tag) =>
-        Stream.eval(config).flatMap(c => AMQP.basicNack(channel, tag, multiple = false, c.requeueOnNack))
+        Stream.eval(F.pure(config)).flatMap(c => AMQP.basicNack(channel, tag, multiple = false, c.requeueOnNack))
```
felixmulder: This could probably be simplified to just what's in the flatMap 😀

core/src/test/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientInMemory.scala
```diff
@@ -47,7 +47,7 @@ class AMQPClientInMemory(internalQ: mutable.Queue[IO, Either[Throwable, AmqpEnve
     val envelope = AmqpEnvelope(DeliveryTag(1), "requeued msg", AmqpProperties.empty)
     for {
       _ <- Stream.eval(ackerQ.enqueue1(NAck(tag)))
-      c <- Stream.eval(config)
+      c <- Stream.eval(IO.pure(config))
```
felixmulder: Same here, this line can be shipped and just put config on the next line 🙂

gvolpe:
Update from Gitter so it doesn't get lost there:

"I would like to propose a different solution: to delegate the responsibility of providing the `Fs2RabbitConfig` with the data to the user. This way we avoid introducing a new library (even though I like `pureconfig` and I currently use it at work) and we get rid of the `"typesafe" config` library. Note, that there's another cool project (https://cir.is/) that users might want to use, so I think as a library giving the decision to the final user is one of the best options."

felixmulder:
Sounds good to me :+1:

insdami:
Cool! I agree, that’s even better I’d say. I’ll get it done when I get to the office 👍

PR #56, opened 2018-03-11
Configuring Travis CI to also build the microsite

gvolpe:
Aggregating the microsite module to be part of the build. This way we can keep documentation updated since for every breaking change the documentation will not compile and will need to be fixed to get a successful build on Travis CI.


insdami:
LGTM 👍

PR #63, opened 2018-05-24
Compile time typesafe arguments based on existential types

### LINKED ISSUE #50: Unsafe arguments
Arguments that are passed through `QueueConfig` could potentially throw a run time exception 
```
ERROR c.g.gvolpe.fs2rabbit.instances.log$ - invalid value in table
java.lang.IllegalArgumentException: invalid value in table
	at com.rabbitmq.client.impl.Frame.fieldValueSize(Frame.java:310) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.Frame.tableSize(Frame.java:250) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.ValueWriter.writeTable(ValueWriter.java:119) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.MethodArgumentWriter.writeTable(MethodArgumentWriter.java:138) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQImpl$Queue$Declare.writeArgumentsTo(AMQImpl.java:1496) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.Method.toFrame(Method.java:85) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQCommand.transmit(AMQCommand.java:104) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQChannel.quiescingTransmit(AMQChannel.java:363) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQChannel.quiescingTransmit(AMQChannel.java:345) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQChannel.quiescingRpc(AMQChannel.java:280) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQChannel.rpc(AMQChannel.java:271) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQChannel.privateRpc(AMQChannel.java:233) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.AMQChannel.exnWrappingRpc(AMQChannel.java:128) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.ChannelN.queueDeclare(ChannelN.java:948) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.rabbitmq.client.impl.recovery.AutorecoveringChannel.queueDeclare(AutorecoveringChannel.java:333) ~[amqp-client-4.1.0.jar:4.1.0]
	at com.github.gvolpe.fs2rabbit.interpreter.AmqpClientStream.$anonfun$declareQueue$1(AmqpClientStream.scala:138) ~[classes/:na]
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12) ~[scala-library.jar:na]
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:74) ~[cats-effect_2.12-0.8.jar:0.8]
	at cats.effect.internals.IORunLoop$.start(IORunLoop.scala:34) ~[cats-effect_2.12-0.8.jar:0.8]
	at cats.effect.IO.unsafeRunAsync(IO.scala:191) ~[cats-effect_2.12-0.8.jar:0.8]
	at cats.effect.IO.$anonfun$runAsync$1(IO.scala:150) ~[cats-effect_2.12-0.8.jar:0.8]
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12) ~[scala-library.jar:na]
	at cats.effect.internals.IORunLoop$.step(IORunLoop.scala:154) ~[cats-effect_2.12-0.8.jar:0.8]
	at cats.effect.IO.unsafeRunTimed(IO.scala:220) ~[cats-effect_2.12-0.8.jar:0.8]
	at cats.effect.IO.unsafeRunSync(IO.scala:173) ~[cats-effect_2.12-0.8.jar:0.8]
	at com.github.gvolpe.fs2rabbit.examples.IOApp.main(IOApp.scala:23) ~[classes/:na]
	at com.github.gvolpe.fs2rabbit.examples.IOApp.main$(IOApp.scala:23) ~[classes/:na]
	at com.github.gvolpe.fs2rabbit.examples.IOAckerConsumer$.main(IOAckerConsumer.scala:26) ~[classes/:na]
	at com.github.gvolpe.fs2rabbit.examples.IOAckerConsumer.main(IOAckerConsumer.scala) ~[classes/:na]
```
Maybe we can create existential types to validate types at compile time using this as a reference
https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/Frame.java#L256

gvolpe:
It closes #50 

Using existential types to provide safe arguments based on the technique described [here](https://www.cakesolutions.net/teamblogs/existential-types-in-scala).

Replaces all the `Map[String, AnyRef]` for a safer version named `Arguments` that checks the types at compile time. This won't compile now for example:

```scala
DeclarationQueueConfig(
  queueName, 
  NonDurable, 
  NonExclusive, 
  NonAutoDelete, 
  Map("key" -> Yolo())
) // type mismatch, expected Arguments
```

However this compiles:
```scala
DeclarationQueueConfig(
  queueName, 
  NonDurable, 
  NonExclusive, 
  NonAutoDelete, 
  Map("key" -> true)
)
```

See `SafeArgsSpec` to see more supported types.

It matches all the types from here: https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/Frame.java#L256 except for `Array` since we support `List` and that's enough.

@insdami let me know what you think.

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
@@ -16,17 +16,22 @@
 
 package com.github.gvolpe.fs2rabbit
 
+import scala.annotation.implicitNotFound
 import scala.language.implicitConversions
-
 import scala.collection.JavaConverters._
 
 object arguments {
 
+  /**
+    * [[SafeArg]] makes sure the arguments passed to any of the functions are compliant with the AMQP Protocol.
+    *
+    * This library only supports String, Boolean, Int, Long, Float, Short, BigDecimal, Date, Byte, List and Map.
+    * */
   type SafeArg   = Evidence[SafeArgument]
   type Arguments = Map[String, SafeArg]
 
   implicit def argumentConversion(arguments: Arguments): java.util.Map[String, Object] =
-    arguments.mapValues(x => x.ev.toObject(x.value)).asJava
+    arguments.map { case (k, v) => k -> v.ev.toObject(v.value) }.asJava
 
   sealed trait Evidence[F[_]] {
     type A
@@ -38,15 +43,16 @@ object arguments {
 
   implicit def anySafeArg[F[_], A: F](a: A): Evidence[F] = MkEvidence(a)
 
-  trait SafeArgument[A] {
+  @implicitNotFound("Only types supported by the AMQP protocol are allowed. Custom classes are not supported.")
+  sealed trait SafeArgument[A] {
     type JavaType >: Null <: AnyRef
     private[fs2rabbit] def toJavaType(a: A): JavaType
     private[fs2rabbit] def toObject(a: A): Object = toJavaType(a)
   }
 
   object SafeArgument {
-    def apply[A](implicit ev: SafeArgument[A]): SafeArgument[A] = ev
-    def instance[A, J >: Null <: AnyRef](f: A => J): SafeArgument[A] =
+    private[fs2rabbit] def apply[A](implicit ev: SafeArgument[A]): SafeArgument[A] = ev
+    private[fs2rabbit] def instance[A, J >: Null <: AnyRef](f: A => J): SafeArgument[A] =
       new SafeArgument[A] {
         type JavaType = J
         def toJavaType(a: A) = f(a)
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2017 Fs2 Rabbit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.gvolpe.fs2rabbit
+
+import scala.language.implicitConversions
+
+import scala.collection.JavaConverters._
+
+object arguments {
+
+  type SafeArg   = Evidence[SafeArgument]
+  type Arguments = Map[String, SafeArg]
+
+  implicit def argumentConversion(arguments: Arguments): java.util.Map[String, Object] =
+    arguments.mapValues(x => x.ev.toObject(x.value)).asJava
+
+  sealed trait Evidence[F[_]] {
+    type A
+    val value: A
+    val ev: F[A]
+  }
+
+  final case class MkEvidence[F[_], A1](value: A1)(implicit val ev: F[A1]) extends Evidence[F] { type A = A1 }
+
+  implicit def anySafeArg[F[_], A: F](a: A): Evidence[F] = MkEvidence(a)
+
+  trait SafeArgument[A] {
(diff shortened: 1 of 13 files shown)


core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import scala.collection.JavaConverters._
+
+object arguments {
+
+  type SafeArg   = Evidence[SafeArgument]
+  type Arguments = Map[String, SafeArg]
+
+  implicit def argumentConversion(arguments: Arguments): java.util.Map[String, Object] =
+    arguments.mapValues(x => x.ev.toObject(x.value)).asJava
```
insdami: I'm not sure whether this still an issue https://stackoverflow.com/questions/25450887/scala-map-mapvalues-stackoverflowerror

core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  trait SafeArgument[A] {
+    type JavaType >: Null <: AnyRef
+    private[fs2rabbit] def toJavaType(a: A): JavaType
+    private[fs2rabbit] def toObject(a: A): Object = toJavaType(a)
+  }
+
+  object SafeArgument {
+    def apply[A](implicit ev: SafeArgument[A]): SafeArgument[A] = ev
+    def instance[A, J >: Null <: AnyRef](f: A => J): SafeArgument[A] =
```
insdami: I'm not sure if it should be public, an user could potential try to 
```scala
case class Yolo()
implicit val yoloInstance: SafeArgument[Yolo] = SafeArgument.instance(identity)
Map("key" -> Yolo())
```

core/src/main/scala/com/github/gvolpe/fs2rabbit/arguments.scala
```diff
@@ -0,0 +1,72 @@
...
+
+package com.github.gvolpe.fs2rabbit
+
+import scala.language.implicitConversions
+
+import scala.collection.JavaConverters._
+
+object arguments {
+
+  type SafeArg   = Evidence[SafeArgument]
```
insdami: This is quite neat, I guess the down-side is that if you tried to 
```scala
DeclarationQueueConfig(..., arguments = Map("key" -> Yolo()))
```
you would get a `type mismatch` saying that `Evidence[SafeArgument]` is expected which can lead to find a way of creating it. My first tough is to customize the error with a `implicitNotFound` which requires an evidence, so I'm not sure if there is a better way to do this

gvolpe: I couldn't come up with a better solution that shows custom `implicit not found` message here (only if I expose the requirement for an implicit in every place the arguments are needed which is not nice). 

For now I documented the purpose of `SafeArg` so people jumping to the definition will understand why it doesn't compile. I sketched some alternative [here](https://gist.github.com/gvolpe/aeb72c0f6a5ba08f491a83399289ad78) but needs to be explored further.

Anyway, I like that it doesn't compile if you pass a custom case class and the user doesn't need to know much about how to create an instance because we must not support that.

PR #67, opened 2018-05-30
Allows client to connect over TLS with the specified SSLContext, if present

kjc1:


gvolpe:
LGTM :+1: 

The documentation is failing to compile, you need to run `sbt` -> `project microsite` -> `tut` to compile the docs (we probably need to document this on the Contribution guide). It'll tell you where it's failing.

Or you can use `sbt` -> `buildFs2Rabbit` but that runs all the `tests` + `coverageReport` + `tut`.

kjc1:
Haha, I had just pushed the fix before you commented! It should build now 🤞

PR #69, opened 2018-06-05
Add deliveryMode and priority to message props

jacemale:
Hey,
First of all, it's a really nice library!

I've added delivery mode and priority to message properties, it required some boxing/unboxing to coerce values to nulls for the Java builder class.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ name := """fs2-rabbit-root"""
 
 organization in ThisBuild := "com.github.gvolpe"
 
-version in ThisBuild := "0.8"
+version in ThisBuild := "0.9-SNAPSHOT"
 
 crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.6")
 
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ name := """fs2-rabbit-root"""
 
 organization in ThisBuild := "com.github.gvolpe"
 
-version in ThisBuild := "0.9-SNAPSHOT"
+version in ThisBuild := "0.8"
 
 crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.6")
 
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -7,7 +7,7 @@ name := """fs2-rabbit-root"""
 
 organization in ThisBuild := "com.github.gvolpe"
 
-version in ThisBuild := "0.8"
+version in ThisBuild := "0.9-SNAPSHOT"
```
gvolpe: Please, refrain from changing the version. This is managed independently :)

core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
```diff
@@ -81,17 +93,21 @@ object model {
...
 
-    def empty = AmqpProperties(None, None, Map.empty[String, AmqpHeaderVal])
+    def empty = AmqpProperties(None, None, None, None, Map.empty[String, AmqpHeaderVal])
 
     def from(basicProps: AMQP.BasicProperties): AmqpProperties =
       AmqpProperties(
         Option(basicProps.getContentType),
         Option(basicProps.getContentEncoding),
+        Option[Integer](basicProps.getPriority).map(Int.unbox),
+        Option(basicProps.getDeliveryMode).map(DeliveryMode.from(_)),
```
gvolpe: Minor nitpick: `.map(DeliveryMode.from)`

gvolpe:
Hi @jacemale , thanks for your PR!

The `CI` is failing because the documentation is not compiling. Run `sbt tut` to compile the docs locally, it's a very small fix :smiley:

gvolpe:
Feel free to ask any questions in the Gitter channel :wink:

core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
jacemale: `DeliveryMode.from` will not compile, because we need an implicit conversion from `Integer` to `Int`

jacemale:
Thanks for review,
I think it should be OK now

PR #76, opened 2018-07-16
Fix bug where headers with arrays would throw a `MatchError` exception.

orium:
Apparently header values can be arrays, in which case fs2-rabbit would throw `MatchError` when trying to convert them to `AmqpHeaderVal`, which prevented it from consuming messages.

Aparently there are other types that are not supported and will cause the same error: https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_3


gvolpe:
Thanks @orium and sorry for the delay in getting back to you, I'm just back from my holidays.

Unfortunately the headers stuff was never fully implemented.

PR #86, opened 2018-09-20
Add support of all standard message properties

### LINKED ISSUE #85: Support all standard properties in AmqpProperties
It looks like there is no way to add a correlation id or a message id for a message. By extension, it would be nice to support all the standard properties (user_id, type, app_id, ...) in addition of those which are already supported (content_type, content_encoding, priority, deliveryMode).

cdelmas:
AmqpProperties previously supported:
- content-type
- content-encoding
- delivery-mode
- priority

This adds support for the following properties:
- correlation-id
- message-id
- type
- app-id
- user-id
- expiration

Additionally, you can still set custom headers in `headers`.

Documentation and examples are up-to-date.

Close #85

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
@@ -96,22 +96,34 @@ object model {
     }
   }
 
-  case class AmqpProperties(contentType: Option[String],
-                            contentEncoding: Option[String],
-                            priority: Option[Int],
-                            deliveryMode: Option[DeliveryMode],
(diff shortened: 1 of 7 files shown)


cdelmas:
It looks like codacy is complaining about using `type` as a field, which makes sensé to me. I should change it for messageType I guess.

gvolpe:
Nevermind Codacy @cdelmas , they have some non-sense rules and I have it as an optional check for PRs. The only mandatory one is the CI build but Travis is randomly failing for no reason but it happened many times before so I'm not surprised.

I just restarted it again, that's the only thing I can do for now, we have to be patient :smile:

PR #124, opened 2018-11-09
First draft of generic AMQP envelope

### LINKED ISSUE #123: Consider making AmqpEnvelope generic
I noticed that the `AmqpMessage` class is generic, but the `AmqpEnvelope` class isn't. Is there a specific reason for that? Because it seems to me that making `AmqpEnvelope` generic would be quite useful, one often needs to pass around the payload together with the metadata (e. g. to Ack or Nack a message later).

gvolpe:
closes #123 

@mberndt123 let me know what you think.

I'm not sure it's worth the pain since it touches many places in the code base and makes the API a bit more complicated. 

My question is when would you need to decode the given payload which is of type `Array[Byte]` into a different type other than a `UTF-8 String`?

Note: Not going to fix the tests until there's a consensus here.


gvolpe:
Closing in favor of #125

PR #125, opened 2018-11-09
[Version 2] - Generic AMQP envelope

### LINKED ISSUE #123: Consider making AmqpEnvelope generic

gvolpe:
Here's a second design of #124  @mberndt123

It only places the `EnvelopeDecoder` constraint on the consumer related methods instead of at the top. I don't normally like to place typeclass constraints on the algebras but this is one way to achieve such goal.

Still the question about when would you need to decode the given payload which is of type `Array[Byte]` into something different than a `UTF-8 String`?

Note: Tests were not fixed either.

closes #123

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
@@ -16,6 +16,8 @@
 
 package com.github.gvolpe.fs2rabbit.effects
 
+import java.nio.charset.StandardCharsets.UTF_8
+
 import cats.ApplicativeError
 
 /**
@@ -32,6 +34,6 @@ object EnvelopeDecoder {
 
   implicit def utf8StringDecoder[F[_]](implicit F: ApplicativeError[F, Throwable]): EnvelopeDecoder[F, String] =
     new EnvelopeDecoder[F, String] {
-      override def decode(raw: Array[Byte]): F[String] = F.catchNonFatal(new String(raw, "UTF-8"))
+      override def decode(raw: Array[Byte]): F[String] = F.catchNonFatal(new String(raw, UTF_8))
     }
 }
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
@@ -18,6 +18,11 @@ package com.github.gvolpe.fs2rabbit.effects
 
 import cats.ApplicativeError
 
+/**
+  * Typeclass that provides the machinery to decode a given AMQP Envelope's payload.
+  *
+  * There's a default instance for decoding payloads into a UTF-8 String.
+  * */
 trait EnvelopeDecoder[F[_], A] {
   def decode(raw: Array[Byte]): F[A]
 }
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
@@ -16,9 +16,11 @@
 
 package com.github.gvolpe.fs2rabbit.effects
 
+import java.nio.charset.Charset
 import java.nio.charset.StandardCharsets.UTF_8
 
 import cats.ApplicativeError
+import com.github.gvolpe.fs2rabbit.model.AmqpProperties
 
 /**
   * Typeclass that provides the machinery to decode a given AMQP Envelope's payload.
@@ -26,7 +28,7 @@ import cats.ApplicativeError
   * There's a default instance for decoding payloads into a UTF-8 String.
   * */
 trait EnvelopeDecoder[F[_], A] {
-  def decode(raw: Array[Byte]): F[A]
+  def decode(raw: Array[Byte], properties: AmqpProperties): F[A]
 }
 
 object EnvelopeDecoder {
@@ -34,6 +36,9 @@ object EnvelopeDecoder {
 
   implicit def utf8StringDecoder[F[_]](implicit F: ApplicativeError[F, Throwable]): EnvelopeDecoder[F, String] =
     new EnvelopeDecoder[F, String] {
-      override def decode(raw: Array[Byte]): F[String] = F.catchNonFatal(new String(raw, UTF_8))
+      override def decode(raw: Array[Byte], properties: AmqpProperties): F[String] = {
+        val encoding = properties.contentEncoding.fold(UTF_8)(Charset.forName)
+        F.catchNonFatal(new String(raw, encoding))
+      }
     }
 }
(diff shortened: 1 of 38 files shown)


mberndt123:
> Still the question about when would you need to decode the given payload which is of type `Array[Byte]` into something different than a `UTF-8 String`?

Well there are plenty of non-UTF-8 formats out there, like MessagePack, BSON, Protobuf etc.. But really the reason that I asked for this is that it's convenient to be able to pass around a piece of data together with the RabbitMQ-related metadata. 
https://www.youtube.com/watch?v=BHjIl81HgfE

core/src/main/scala/com/github/gvolpe/fs2rabbit/effects/EnvelopeDecoder.scala
```diff
@@ -0,0 +1,37 @@
...
+trait EnvelopeDecoder[F[_], A] {
+  def decode(raw: Array[Byte]): F[A]
+}
+
+object EnvelopeDecoder {
+  def apply[F[_], A](implicit ev: EnvelopeDecoder[F, A]): EnvelopeDecoder[F, A] = ev
+
+  implicit def utf8StringDecoder[F[_]](implicit F: ApplicativeError[F, Throwable]): EnvelopeDecoder[F, String] =
+    new EnvelopeDecoder[F, String] {
+      override def decode(raw: Array[Byte]): F[String] = F.catchNonFatal(new String(raw, "UTF-8"))
```
mberndt123: You can use `java.nio.charset.StandardCharsets.UTF_8` here instead of a string literal

gvolpe:
> Well there are plenty of non-UTF-8 formats out there, like MessagePack, BSON, Protobuf etc.. But really the reason that I asked for this is that it's convenient to be able to pass around a piece of data together with the RabbitMQ-related metadata.

I'm aware of the different formats :) I just found out mostly everyone uses `UTF-8` and I was wondering whether these changes were worth it because it's not simply adding a type parameter. But if you think it might be useful for you that's enough to move forward with these PR.

gvolpe:
It should be ready to go now :+1: 

@mberndt123 since you're getting involved quite a lot would you like to become a maintainer? The project could definitely use more help.

mberndt123:
Hi @gvolpe,

> It is just a typeclass, not a monad transformer. What makes you think that? In order to call `flatMap` you need to have at least `FlatMap[F]`. Same shape as `ApplicativeError[F, E]` for example.

Well, in Scala these aren't mutually exclusive. As far as the language is concerned, a type class is just some generic type that is usually passed around implicitly, and a monad transformer is a generic type that, when partially applied to a monad, yields another monad. A type can easily fit both definitions, and `EnvelopeDecoder` in fact does. It's easy to write a definition of `Monad[EnvelopeDecoder[F, ?]]`, hence it is a monad transformer. Take a look at the definitions:
```scala
trait EnvelopeDecoder[F[_], A] {
  def decode(raw: Array[Byte], properties: AmqpProperties): F[A]
}
```
```scala
final case class Kleisli[F[_], A, B](
  run: A => F[B]
)
```
(`ReaderT` is just an alias for `Kleisli`)
If you substitute `A = (Array[Byte], AmqpProperties)` in the definition of Kleisli you end up with the same result (modulo naming). So I don't really see the point in adding a new trait when Kleisli does exactly the same thing.

> I strongly disagree on using Monad Transformers when they can be avoided unless we have a strong usage case.

This debate isn't about whether or not we're going to use monad transformers – we'll  do that either way, because `EnvelopeDecoder` is a monad transformer just like `Kleisli`. The debate is about whether we'll use a custom implementation or just use the one that's there already in cats.

mberndt123:
Hey @gvolpe,


I think we've been talking past each other a bit, I should probably explain a bit better what I'm trying to do here. I'm not proposing to change the monad that users write their code in. People would still write their code in the `IO` monad (or `Task` or whatever), and `EnvelopeDecoder`s would still be a type class that is passed around implicitly, just like in your proposal. So there will be no additional allocations or anything like that. 

The point here is that I'd like to make it easier to create new instances of `EnvelopeDecoder`. For instance, given an `EnvelopeDecoder[F, A]` and a function `A => B`, I should be able to create an `EnvelopeDecoder[F, B]`. In other words, we need a `Functor` instance for `EnvelopeDecoder`:
```scala
implicit def envelopeDecoderFunctor[F[_]: Functor] = new Functor[EnvelopeDecoder[F[_], ?]] {
  def map[A](fa: EnvelopeDecoder[F, A])(f: A => B) = new EnvelopeDecoder[F, B] {
    def decode(bytes: Array[Byte], props: AmqpProperties) =
      fa(bytes, props).map(f)
  }
}
```
Now, assuming a `case class Foo(s: String)`, I can define an `EnvelopeDecoder` for it using 
```scala
implicit def fooDecoder[F[_]](implicit stringDecoder: EnvelopeDecoder[F, String], F: Functor[F]) =
  stringDecoder.map(Foo.apply)
```
Surely that's much nicer than writing a new `EnvelopeDecoder` from scratch!

And there are also cases where you not only need `Functor` but also `Monad` or `ApplicativeError`. It's easily possible to implement all these type classes for your `EnvelopeDecoder`, but we don't have to: it's all already there in `Kleisli`. Again, the only purpose here is to give users an easy way to define `EnvelopeDecoder`s by combining existing ones in a simple but flexible way. Let's assume my `Kleisli` based implementation for `EnvelopeDecoder` for a second:

```scala
type EnvelopeDecoder[F[_], A] = ReaderT[F, AmqpEnvelope[Array[Byte]], A]
```
Now I can easily define a decoder to extract just the encoding:
```scala
def encoding[F[_]](implicit F: ApplicativeError[F, Throwable]): EnvelopeDecoder[F, Option[Charset]] =
 Kleisli(_.properties.contentEncoding.traverse(n => ApplicativeError[EnvelopeDecoder[F, ?]].catchNonFatal(Charset.forName(n))))
```
and for the bytes:
```scala
def bytes[F[_]: Applicative]: EnvelopeDecoder[F, Array[Bytes]] =
  Kleisli(_.payload.pure[F])
```
And I can easily combine them:
```scala
implicit def stringDecoder[F[_]: MonadError]: EnvelopeDecoder[F, String] =
  for {
    charset <- encoding
    b <- bytes
  } yield new String(b, charset.getOrElse(UTF_8))
```
(I might have gotten some of the details wrong here since I didn't run this through a compiler, but it does work this way)

gvolpe:
Hi @mberndt123 ,

Thanks for the detailed example. We were indeed discussing different things and I replied to quick without seeing your actual implementation in the other PR where you literally call `decoder.run(...)` which is equivalent to invoke a function (minus the allocation of `Kleisli`).

I think you made a good point about composition and ease to derive a new instance of `EnvelopeDecoder[F, A]`. After all the idea is to give the user easier and better tools :)

So just to move forward, because it's hard to keep up with the timezone difference + I'm currently out of the country for a conference, I'll get this PR merged and afterwards get yours merged with your changes okay?

PR #127, opened 2018-11-10
replace EnvelopeDecoder with ReaderT

mberndt123:
Hey,

I've implemented the changes I've suggested [here](https://github.com/gvolpe/fs2-rabbit/pull/125#issuecomment-437580203).

I managed to fix the implicit lookup issue I mentioned there using tagged types, I'll be glad to explain the details if needed. I used a library for the tagging, but it's less than 10kb in size, so I don't think it's a problem.

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
@@ -16,11 +16,18 @@
 
 package com.github.gvolpe.fs2rabbit
 
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets.UTF_8
+
+import cats.ApplicativeError
+import cats.data.Kleisli
 import com.github.gvolpe.fs2rabbit.arguments.Arguments
+import com.github.gvolpe.fs2rabbit.effects.EnvelopeDecoder
 import com.github.gvolpe.fs2rabbit.model.AmqpHeaderVal._
 import com.rabbitmq.client.impl.LongStringHelper
 import com.rabbitmq.client.{AMQP, Channel, LongString}
 import fs2.{Sink, Stream}
+import cats.implicits._
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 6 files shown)


mberndt123:
I've removed the dependency because it turned out to be unnecessary; the UTF-8 `EnvelopeDecoder` can simply be put in the `AmqpEnvelope` companion object instead. See the other PR for a discussion about `EnvelopeDecoder` vs. `Kleisli`/`ReaderT`.

core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
```diff
@@ -154,6 +159,12 @@ object model {
   case class AmqpEnvelope[A](deliveryTag: DeliveryTag, payload: A, properties: AmqpProperties)
   case class AmqpMessage[A](payload: A, properties: AmqpProperties)
 
+  object AmqpEnvelope {
```
gvolpe: I'd prefer to have this default instance in the companion object of `EnvelopeDecoder`. I think it's good practice to expose a summoner method.

```scala
object EnvelopeDecoder {
  def apply[F[_], A](implicit ev: EnvelopeDecoder[F, A]): EnvelopeDecoder[F, A] = ev
}
```

Other than that looks good!

mberndt123: `object EnvelopeDecoder` is unfortunately not the companion object to the `EnvelopeDecoder` type since that is only a type alias now, so the companion object is `Kleisli`. I can still put type class instances in `object EnvelopeDecoder`, but then they wouldn't be found by the implicit lookup. So I'd rather keep them in the `AmqpEnvelope` object as they will be found there.

I will add a summoning method. I'd also like to add a few more helper functions that people are likely to need.

mberndt123:
I'd say this is done.

By the way, the automatically inserted license header contains a 2017 date, perhaps it's time to update that

core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
gvolpe: Yes, I meant just the `EnvelopeDecoder` object.

> So I'd rather keep them in the AmqpEnvelope object as they will be found there.

Are you sure that would work? I believe that you will need to import `AmqpEnvelope._` to access the instances right? If that's the case I believe it's better to have them in the `EnvelopeDecoder` object and import that instead.

gvolpe:
Cool I just raised an issue for the licence headers. 

Only missing the docs, if you can do it that'd be great otherwise tomorrow I'll have some time to tackle it: https://github.com/gvolpe/fs2-rabbit/blob/master/site/src/main/tut/consumers/index.md

core/src/main/scala/com/github/gvolpe/fs2rabbit/model.scala
mberndt123: No, you don't have to import anything. When an implicit of type `Foo[Bar]` is needed, the compiler will search the companion objects of both `Foo` and `Bar`, and since `AmqpEnvelope` is in fact used in `Kleisli[F, AmqpEnvelope[Array[Byte]], A]` its companion object will be searched.

gvolpe: Wow, I wouldn't have expected that. Implicits can be obscure sometimes :sweat_smile:

mberndt123:
I've pushed some documentation about this. I haven't checked to see if the code in the examples actually compiles, but it would be nice to get tut to do this, right? Does the CI job do that?

PR #131, opened 2018-11-17
improve error handling for envelope decoder failures

mberndt123:
Errors in the EnvelopeDecoder aren't currently handled at all, this patch should improve things a bit.

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientStream.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientStream.scala
@@ -18,7 +18,6 @@ package com.github.gvolpe.fs2rabbit.interpreter
 
 import cats.effect.Effect
 import cats.effect.syntax.effect._
-import cats.syntax.flatMap._
 import com.github.gvolpe.fs2rabbit.algebra.{AMQPClient, AMQPInternals}
 import com.github.gvolpe.fs2rabbit.arguments._
 import com.github.gvolpe.fs2rabbit.config.declaration.{DeclarationExchangeConfig, DeclarationQueueConfig}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientStream.scala
```diff
@@ -29,6 +28,7 @@ import com.github.gvolpe.fs2rabbit.effects.BoolValue.syntax._
 import com.github.gvolpe.fs2rabbit.effects.{EnvelopeDecoder, StreamEval}
 import com.rabbitmq.client._
 import fs2.Stream
+import cats.implicits._
```
gvolpe: Let's use just `cats.syntax.all._` or being more picky just add `import cats.syntax.applicativeError._`

I know this is a matter of preference but I tend not to use `import cats.implicits._` to speed up compilation time and to be kind with those poor souls using IJ Idea :)

mberndt123: I'd rather not make that change. Apparently this code base is compiled for Scala 2.11 and 2.12. I'm calling `map` on an `Either`, and since pre-2.12 `Either` isn't right-biased, that doesn't work unless I also import `cats.instances.either._`. However because it is right-biased in 2.12 I'll now get an unused import error. All solutions to that problem make the code uglier, and for what?
People who have studied the issue have told me that using only specific imports doesn't actually improve compile times to a meaningful degree.

By the way, I'm one of the poor souls using IntelliJ. I don't know of a better alternative (Ensime isn't one, and it's pretty much dead anyway).

gvolpe: The issue with Scala 2.11 is a very strong argument and I agree with it. The second part I think is a matter of preference :) 

So cool, go for it but could you please organize the imports? I kinda have OCD :laughing:

mberndt123: I don't understand, what do you want me to do here? Arrange imports alphabetically?

PR #132, opened 2018-11-17
Simplify StreamAcker and StreamPublisher

mberndt123:
Hi @gvolpe,

I've gained some experience with the library and I think that defining `StreamAcker` as a `Sink` is not the best possible choice. For instance, suppose you want to transform the stream of messages to some other stream and acknowledge the results along the way. With the current design, you have to do something like this: 
```scala
val consumer: StreamConsumer[F, A] = ???
val acker: StreamAcker[F] = ???
consumer.flatMap { a =>
  Stream.eval(doStuff(a)).flatMap { result =>
    Stream.emit(Ack(a.deliveryTag)).to(acker).drain ++ Stream.emit(result)
}
```

With my new definition, this is more straight-forward:
```scala
consumer.evalMap { a =>
  doStuff(a).flatMap { result =>
    acker(AckResult(a.deliveryTag)).as(result)
}
```
I feel like doing things in terms of `Stream[F, ?]` rather than `F` doesn't really buy you anything in this case and makes things more complicated (and inefficient, because it breaks fs2's chunking) instead. I also think it's a bit error-prone, because people are likely to `flatMap` over the `Stream[F, Unit]` that `.to(acker)` returns (instead of using `++`). If some day we change the implementation to return e. g. an empty `Stream[F, Unit]` instead of a `Stream[F, Unit]` containing a single `Unit` value, people's programs will break. I've seen people make that mistake too often to think that it's not going to happen :wink: 

Existing programs are trivial to adapt, because they probably just need to change from `.to(acker)` to `.evalMap(acker)`, as shown in the example programs.

Let me know what you think about the proposal.

I also feel that a similar reasoning can be applied to `StreamPublisher` (actually even more so!), so I'd like to propose a similar change for `StreamPublisher` too. But first let me know what you think!

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientStream.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientStream.scala
@@ -113,7 +113,7 @@ class AMQPClientStream[F[_]: Effect](implicit SE: StreamEval[F]) extends AMQPCli
       exchangeName: ExchangeName,
       routingKey: RoutingKey,
       msg: AmqpMessage[String]
-  ): F[Unit] = Effect[F].delay {
+  ): F[Unit] = Sync[F].delay {
     channel.basicPublish(
       exchangeName.value,
       routingKey.value,
@@ -128,7 +128,7 @@ class AMQPClientStream[F[_]: Effect](implicit SE: StreamEval[F]) extends AMQPCli
       routingKey: RoutingKey,
       flag: PublishingFlag,
       msg: AmqpMessage[String]
-  ): F[Unit] = Effect[F].delay {
+  ): F[Unit] = Sync[F].delay {
     channel.basicPublish(
       exchangeName.value,
       routingKey.value,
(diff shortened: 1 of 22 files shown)


mberndt123:
OK, I've made the same change for `StreamPublisher`, which is now defined as `AmqpMessage[String] => F[Unit]`.

This makes it easier to use the library in a context without streams (e. g. an http4s app  where you need to return an `F[Response]`), but also in a streaming context (e. g. you can use `evalTap` instead of `observe`, which is much more efficient).

core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/AMQPClientStream.scala
```diff
@@ -113,7 +113,7 @@ class AMQPClientStream[F[_]: Effect](implicit SE: StreamEval[F]) extends AMQPCli
       exchangeName: ExchangeName,
       routingKey: RoutingKey,
       msg: AmqpMessage[String]
-  ): Stream[F, Unit] = SE.evalDiscard {
+  ): F[Unit] = Effect[F].delay {
```
gvolpe: Prefer to use `Sync[F]` here?

gvolpe:
Let me know if you have any other proposals that make sense before `1.0.0`. You can join the Gitter channel for a quick chat. I'd like to plan the release of the first stable version so maybe other changes can be postponed for a future `1.1.x`.

PR #133, opened 2018-11-21
Improve names

ksonj:
This is an attempt at finding a more consistent naming scheme. Specifically this introduces the following changes:

- `StreamPublisher` is renamed to `Publisher` (there is no `Stream` in its definition)
- likewise `StreamAcker` is renamed to `Acker`
- and the previously existing trait `Acker` is renamed to `Acking` (to be in the same grammatical form as `Publishing` and `Consuming`)
- For the same reasons the trait `Consumer` is renamed to `Consuming`
- and the previously existing trait `Consuming` is renamed to `AckConsuming` (which probably isn't a great name, but yet imo better communicates its intent)
- Finally the implementing programs `AckerProgram`, `ConsumerProgram` and `ConsumingProgram` are renamed to match (i.e. `AckerProgram` -> `AckingProgram`, `ConsumerProgram` -> `ConsumingProgram` and `ConsumingProgram` -> `AckConsumingProgram`)


mberndt123:
I agree with ksonj that we don't really need the `Publisher` and `Acker` aliases. I kept them around because that issue is orthogonal to what I've been working on, but I also find it easier to read the code without having to look up what those types are.

gvolpe:
Sounds good to me @ksonj , :+1: to remove the type aliases.

PR #136, opened 2018-11-21
Extend publishing algebra to delay the decision for a routing key

ksonj:
I'd like to use this PR as a basis to discuss the introduction of some means to dynamically decide on routing keys for publishing.

This iteration proposes the introduction of two new methods `createRoutingPublisher` and `createRoutingPublisherWithListener` to the `Publishing`-algebra which
add the ability to create a `RoutingKey => Publisher[F]` (i.e. `RoutingKey => AmqpMessage[String] => F[Unit]`) to defer the decision for a routing key to the call site.

Another possibility that comes to mind is adding a `type RoutedAmqpMessage[A] = (RoutingKey, AmqpMessage[A])` and return `RoutedAmqpMessage[String] => F[Unit]`

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Publishing.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Publishing.scala
@@ -36,4 +36,16 @@ trait Publishing[F[_], G[_]] {
       listener: PublishReturn => G[Unit]
   )(implicit encoder: MessageEncoder[G, A]): F[A => G[Unit]]
 
+  def createRoutingPublisher[A](
+      channel: Channel,
+      exchangeName: ExchangeName
+  )(implicit encoder: MessageEncoder[G, A]): F[RoutingKey => A => G[Unit]]
+
+  def createRoutingPublisherWithListener[A](
+      channel: Channel,
+      exchangeName: ExchangeName,
+      flags: PublishingFlag,
+      listener: PublishReturn => G[Unit]
+  )(implicit encoder: MessageEncoder[G, A]): F[RoutingKey => A => G[Unit]]
+
 }
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/Fs2Rabbit.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/Fs2Rabbit.scala
@@ -89,6 +89,18 @@ class Fs2Rabbit[F[_]: Concurrent] private[fs2rabbit] (
   )(implicit channel: AMQPChannel, encoder: MessageEncoder[F, A]): Stream[F, A => F[Unit]] =
     publishingProgram.createPublisherWithListener(channel.value, exchangeName, routingKey, flags, listener)
 
+  def createRoutingPublisher[A](exchangeName: ExchangeName)(
+      implicit channel: AMQPChannel,
+      encoder: MessageEncoder[F, A]): Stream[F, RoutingKey => A => F[Unit]] =
+    publishingProgram.createRoutingPublisher(channel.value, exchangeName)
+
+  def createRoutingPublisherWithListener[A](
+      exchangeName: ExchangeName,
+      flags: PublishingFlag,
+      listener: PublishReturn => F[Unit]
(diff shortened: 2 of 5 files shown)


gvolpe:
Maybe better to get #133 merged before this one?

gvolpe:
@ksonj sync this branch with master and it'll be ready to be merged :)

ksonj:
@gvolpe Nice, I'll see that I get to that tomorrow. What's your preference with history? Should I rebase here or make a new PR?

gvolpe:
Cool thanks @ksonj ! Just rebase here :+1:

mberndt123:
`RoutingKey => AmqpMessage[String] => F[Unit]` is isomorphic to `(RoutingKey, AmqpMessage[String]) => F[Unit]`, and I'd argue the latter is more idiomatic in Scala.

ksonj:
@mberndt123 Alright, either's fine for me. Although I find the curried thing is at least nicer to read. As to what's more idiomatic I don't know, I'm fairly new to the Scala game. Let's hear what @gvolpe has to say and then I'll update if necessary

gvolpe:
I agree that using a tuple is probably more standard but in this case I'm in favor of the curried version since it makes usability much nicer. Eg: we can now do this:

```scala
msg.covary[IO] evalMap publisher(routingKey)
```

If it was defined as `(RoutingKey, AmqpMessage[String]) => F[Unit]` we would have to use it as:

```scala
msg.covary[IO].evalMap(msg => publisher(routingKey -> msg))`
```

mberndt123:
> I agree that using a tuple is probably more standard

There is no tuple there. `(A, B) => C` is syntactic sugar for `Function2[A, B, C]`, not for `Function1[Tuple2[A, B], C]`. A function over tuples is written `((A, B)) => C`). 


> If it was defined as `(RoutingKey, AmqpMessage[String]) => F[Unit]` we would have to use it as:
> 
> ```scala
> msg.covary[IO].evalMap(msg => publisher(routingKey -> msg))`
> ```

No, it would be
```scala
msg.covary[IO].evalMap(publisher(routingKey, _))
```

I also suspect that this is not a very common case. After all the whole point here is *not* to use the same routingKey for all the messages. And the (IMO more common) case of specifying both is a bit more awkward with the curried version: `publisher(routingKey)(msg)`. So yeah, I pretty much still think that a binary function type would be more appropriate here. I don't think it's very important though, so if you still like the curried version better, go ahead.

mberndt123:
Oh, and now that #137 is merged to master, the type should probably be `(RoutingKey, A) => F[Unit]`

ksonj:
@mberndt123 You're making a good point. I was reading that as a tuple as well, tbh. Though the distinction doesn't strike me as being super important here. `((A, B)) => C` should also be isomorphic to `(A, B) => C`, right?

Yet I remain indifferent. If it weren't for all the brackets, I'd always prefer a curried function, but I get that that isn't very practical in general. And possibly has other drawbacks on the JVM that I might be totally unaware of. So your call @gvolpe 

> Oh, and now that #137 is merged to master, the type should probably be (RoutingKey, A) => F[Unit]

It's already `A` instead of `AmqpMessage[String]` in this PR

mberndt123:
> @mberndt123 You're making a good point. I was reading that as a tuple as well, tbh. Though the distinction doesn't strike me as being super important here. `((A, B)) => C` should also be isomorphic to `(A, B) => C`, right?

Yes, that's right. 

> Yet I remain indifferent. If it weren't for all the brackets, I'd always prefer a curried function, but I get that that isn't very practical in general. And possibly has other drawbacks on the JVM that I might be totally unaware of. So your call @gvolpe

Well one obvious drawback is that it leads to more allocations, though that's probably irrelevant here.

> It's already `A` instead of `AmqpMessage[String]` in this PR

Oh right, fine then!

gvolpe:
Thanks for your insights!

Let's leave it like it is then. It is not a common use case anyway so I think we had enough discussions but I'd be happy to open a different issue / PR if we discover something can be improved. 

I believe `msg.covary[IO] evalMap publisher(routingKey)` is still more idiomatic than `msg.covary[IO].evalMap(publisher(routingKey, _))` but that's just a matter of preference.

PR #142, opened 2018-11-27
Handle shutdown event of an acquired connection

iRevive:
After shutdown event an acquired connection will be terminated

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/ConnectionStream.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/ConnectionStream.scala
@@ -18,7 +18,9 @@ package com.github.gvolpe.fs2rabbit.interpreter
 
 import javax.net.ssl.SSLContext
 
-import cats.effect.Sync
+import cats.effect.concurrent.MVar
+import cats.effect.syntax.effect._
+import cats.effect.{ConcurrentEffect, Sync}
 import cats.syntax.apply._
 import cats.syntax.flatMap._
 import cats.syntax.functor._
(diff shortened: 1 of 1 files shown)


gvolpe:
I thought about this and I think this could made opt-in but I would prefer to keep the current behavior as the standard, which is provided by the underlying Java driver.

I believe most people don't really care about manually taking care of connection failures if the library can do it for them. Eg: This is what happens when running the `IOAckerConsumer` example followed by shutting down the `RabbitMQ` server and starting it up again a few minutes later:

```
Consumed: AmqpEnvelope(DeliveryTag(1),Hey!,AmqpProperties(None,None,None,None,None,None,None,None,None,None,Map(app -> StringVal(fs2RabbitDemo), demoId -> LongVal(123))))
Consumed: AmqpEnvelope(DeliveryTag(2),{"id":1,"name":"Sherlock","address":{"number":212,"streetName":"Baker St"}},AmqpProperties(None,None,None,None,None,None,None,None,None,None,Map()))

-- Shutdown server
17:50:59.627 [AMQP Connection 127.0.0.1:5672] WARN  c.r.c.impl.ForgivingExceptionHandler - An unexpected connection driver error occured (Exception message: Connection reset)
17:50:59.627 [RabbitMQ Error On Write Thread] WARN  c.r.c.impl.ForgivingExceptionHandler - An unexpected connection driver error occured (Exception message: Connection reset)

-- Restart server
-- New message published on the queue
Consumed: AmqpEnvelope(DeliveryTag(3),test,AmqpProperties(None,None,None,Some(NonPersistent),None,None,None,None,None,None,Map()))
```

gvolpe:
By opt-in I mean we could probably add an extra parameter to the configuration.

iRevive:
If reconnection works automatically under the hood, then this functionality is not really necessary.
I agree with you, it should be a part of the configuration.

gvolpe:
Re-connection always worked. You don't see a message of the connection being released because it is still being used and actively trying to reconnect. This is done by the Java driver, that's why there are no logs about it.

gvolpe:
Closing for now. Feel free to re-open or create a different PR with the opt-in solution.

PR #150, opened 2018-12-13
[WIP] Add an implementation in terms of cats-effect only

ksonj:
This is a WIP-attempt at simplifying the library by reimplementing the API in terms of cats-effect instead of `Stream`. IMO this has several advantages:

- the implementation becomes simpler (and there less error prone and easier to maintain and reason about)
- it enables use of this library in scenarios where the power of streaming is not needed or desired
- at the same time we do not lose any power, as (I believe) the current API can easily be reconstructed on top of this

I think the biggest user-facing change in the end is that a "consumer" is now represented by `F[AmqpEnvelope[A]]` which can trivially be turned back into a `Stream[F, AmqpEnvelope[A]]` via `repeatEval` by the user. The other obvious change is that instead of `Stream.bracket` we're using `cats.effect.Resource` to manage the connection. Again, this can easily be changed on demand via `Stream.resource`. Generally all operations that before returned singleton streams simply return `F` now.
A look at the two mini-examples maybe illustrates that.

This has been a very enjoyable tour through `cats.effect` and `fs2` so far, so even if this turns out to be a bad idea or too drastic of a change to get accepted, I've definitely learned a lot along the way. In any case I'd be very happy if @gvolpe and maybe @mberndt123 could give some feedback on this. Is there anything obviously broken about this approach?

There's obviously still a few things that'd need fixing, e.g. I still need to reimplement the previously existing examples to prove that all that still works as originally intended. Also, I haven't really looked at the circe stuff yet and there's quite likely yet a few things that need improvement or do not even work. I thought I get this out anyway in hope of early feedback. Obviously the guide would need to be adapted as well


.gitignore
```diff
@@ -48,3 +48,5 @@ derby.log
 /work
 /${json:qpid.work_dir}
 tags
+.metals/
```
gvolpe: This is okay because I also use `metals` but for next time prefer to use a global `gitignore` file for custom things :)

ksonj:
Nice!
> Anyway I'd like to publish a 1.0.0 before this change goes in as it is quite significant so we are not going to merge this immediately.

yes absolutely, there's still some work left to do before this is back on par anyway

.gitignore
ksonj: :+1: that somehow slipped through :slightly_smiling_face:

core/src/main/scala/com/github/gvolpe/fs2rabbit/program/ConsumingProgram.scala
```diff
@@ -39,14 +35,12 @@ class ConsumingProgram[F[_]: MonadError[?[_], Throwable]](AMQP: AMQPClient[Strea
...
-      internalQ <- Stream.eval(IQ.create)
+      internalQ <- IQ.create
       internals = AMQPInternals[F](Some(internalQ))
       _         <- AMQP.basicQos(channel, basicQos)
       consumeF  = AMQP.basicConsume(channel, queueName, autoAck, consumerTag, noLocal, exclusive, args)(internals)
-      _         <- Stream.bracket(consumeF)(tag => AMQP.basicCancel(channel, tag))
-      consumer <- Stream.repeatEval(
-                   internalQ.dequeue1.rethrow.flatMap(env => decoder(env).map(a => env.copy(payload = a))))
-    } yield consumer
+      resource  = Resource.make(consumeF)(tag => AMQP.basicCancel(channel, tag))
```
ksonj: @gvolpe I'll have to revisit this. This currently sends basicCancel for every message, which obviously isn't intended. The signature of createConsumer probably has to change to include `Resource` as well.

core/src/test/scala/com/github/gvolpe/fs2rabbit/interpreter/Fs2RabbitSpec.scala
```diff
@@ -66,36 +67,19 @@ class Fs2RabbitSpec extends FlatSpecLike with Matchers {
...
-    * */
-  def rabbitRTS(
-      ref: Ref[IO, AMQPInternals[IO]],
-      publishingQ: Queue[IO, Either[Throwable, AmqpEnvelope[Array[Byte]]]]
-  ): Stream[IO, Unit] =
-    Stream.eval(ref.get).flatMap { internals =>
-      internals.queue.fold(rabbitRTS(ref, publishingQ)) { internalQ =>
-        for {
-          _ <- Stream.eval(publishingQ.dequeue1.flatMap(internalQ.enqueue1))
-          _ <- Stream.eval(ref.set(AMQPInternals(None)))
```
ksonj: What is this needed for?

gvolpe:
Hi @ksonj thanks for taking the time to do this. I went through the code changes and I'm not convinced this is the way forward. 

I think that many parts of the API can be expressed in terms of `F[A]` but there's one fundamental part where I believe we should keep the `Stream[F, A]` interface and this is the `Consumer`. The idea from the beginning was to wrap the underlying callback-y interface given by the Java client and a RabbitMQ consumer is normally a long-running computation better expressed with streams. Although, as you mentioned, we could probably express the same API by doing a `Stream.repeatEval(Stream.resource(foo))` I'm convinced this should be provided by the library and I don't really see the benefit in exposing the underlying `Resource`.

Not saying that we should dump the work you have done. I believe we can take a lot of it but I'd prefer to maintain the stream-based API at least for the consumer. Probably the rest can be expressed just as an effect `F[A]`. What do you think?

ksonj:
Yes, you're probably right about the consumer. After all I don't really see why one would not want to use it that way. I'll see that I get back to this after all the christmas and new year's stuff :) happy holidays until then!

gvolpe:
Closing in favor of #189 . Thanks again for all the work done here @ksonj , feel free to chime in :)

ksonj:
Thanks @gvolpe, and sorry for letting this rot for so long

gvolpe:
No worries, it was a tough one and we all have a life :wink:

PR #151, opened 2018-12-16
nack messages that can't be decoded

mberndt123:
Hi,

it's currently quite hard to NAck messages when the envelope decoder fails to decode them. But I think that's basically always what you want, so I propose to do it in the library. People who don't want that can always use an `EnvelopeDecoder[F, Either[Throwable, A]]` and do all the error handling manually.

I haven't bothered fixing the tests, docs or examples, I'll do that later if we can agree on this behaviour.

Cheers,
Matthias

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Consuming.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Consuming.scala
@@ -18,7 +18,7 @@ package com.github.gvolpe.fs2rabbit.algebra
 
 import com.github.gvolpe.fs2rabbit.arguments.Arguments
 import com.github.gvolpe.fs2rabbit.effects.EnvelopeDecoder
-import com.github.gvolpe.fs2rabbit.model.{AmqpEnvelope, BasicQos, ConsumerTag, QueueName}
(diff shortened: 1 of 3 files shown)


ksonj:
I think that's sensible. I wonder if this would become simpler with something like what we've talked about over the weekend on gitter, i.e. changing the way acking is handled in general. https://gitter.im/fs2-rabbit/fs2-rabbit?at=5c1520ff539aca60b2149d70

mberndt123:
Hey @ksonj,

I agree that simplified acking is desirable, and it would probably simplify the implementation of this feature a bit. However I also think that these two are unrelated when it comes to the functionality, and I would like to be able to move this PR forward independently of the other change.

Cheers,
Matthias

gvolpe:
Yes, I don't know what I was thinking. It doesn't make sense in automatic `ack` mode...

> I think it should be the default.

I agree. Let's just give the user the option to deactivate it though.

> So maybe the best thing to do would be to just throw an Exception like this if the decoder fails.

That makes sense. You could use something like `.adaptError { case NonFatal(e) => DecodingError(env, e) }` there and it'll make sense whether the auto-nacking is active or not.

gvolpe:
MIght come back to it at some point if anyone is interested in this functionality :)

PR #157, opened 2019-01-16
Provide an option to configure multiple nodes of a cluster

### LINKED ISSUE #155: Failover to next node if current node becomes unavailable
Hey @gvolpe ,
while evaluating your very promising library for use in a new service, I noticed that specifying multiple nodes of the same RabbitMQ cluster does not seem to be supported. The expectation would be to have the service automatically establish a connection to node B in case node A becomes unreachable, if both, node A and B are specified in the configuration.

I tried to work around this by defining multiple `Fs2RabbitConfig`s, instantiating one `Fs2Rabbit[F]` for each of them and then calling my own `createConnectionChannel()` to pick the first successful connection:

```scala
object Fs2RabbitUtil {

  private val logger = getLogger

  def createConnectionChannel[F[_]: Sync](
      nodes: List[Fs2Rabbit[F]]
  ): Stream[F, (Fs2Rabbit[F], AMQPChannel)] =
    nodes match {
      case node :: otherNodes =>
        (
          for {
            channel <- node.createConnectionChannel
            _       <- Stream.eval(Sync[F].delay(logger.info("Connected to RabbitMQ node.")))
          } yield (node, channel)
        ) handleErrorWith {
          case NonFatal(e) =>
            for {
              _       <- Stream.eval(Sync[F].delay(logger.warn(e)(s"Connection attempt to RabbitMQ node failed.")))
              channel <- createConnectionChannel(otherNodes)
            } yield channel
        }
      case List() => Stream.raiseError[F](new Exception(s"Connection attempts to all RabbitMQ nodes failed."))
    }
}
```

In connection with `ResilientStream.run()` I would have expected this to be enough to fail the stream in case a node becomes unreachable and then `createConnectionChannel` would pick the next node and restart the stream. However `amqp-client` only knows about one node and since [automatic recovery is not disabled](https://github.com/gvolpe/fs2-rabbit/blob/deeeae432a3f2c4293307fcb1ca99c5d4fe79233/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/ConnectionStream.scala#L64), `amqp-client` will try to reconnect to the same node over and over again instead of letting the stream fail.

There's two possible solutions to support failover that I can think of:
- Disable automatic recovery in `amqp-client`'s `ConnectionFactory` and let the stream fail. Users of the library would then have to handle the failover themselves.
- Support multiple nodes in `Fs2RabbitConfig` and pass them on to `amqp-client`, e.g.:
    ```scala
    Fs2RabbitConfig(
        nodes = List(
            Fs2RabbitNodeConfig(
                host = "host1.example.net",
                port = 5672
            ),
            Fs2RabbitNodeConfig(
                host = "host2.example.net",
                port = 5672
            )
        ),
        virtualHost = "/",
        connectionTimeout = 10,
        ssl = false,
        username = None,
        password = None,
        requeueOnNack = true,
        internalQueueSize = None
     )
    ```
    `amqp-client` [already supports configuring multiple nodes](https://www.rabbitmq.com/api-guide.html#address-array) and [will fall back to the other nodes automatically](https://www.rabbitmq.com/api-guide.html#connection-recovery).

JanGe:
Also expose the `automaticRecovery` option of `amqp-client`.

Fixes #155 .

Caveats:
- It's not really backwards compatible, since e.g. `pureconfig` does not recognize the new `apply` method with the old parameter list and also a `config.copy(port=1234)` wouldn't compile now, but compiled before.
- Since I used a `NonEmptyList`, `pureconfig` users would now also have to import `pureconfig-cats`. I can change it to use a normal `List`, if you prefer that @gvolpe .

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/config/Fs2RabbitConfig.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/config/Fs2RabbitConfig.scala
@@ -16,14 +16,46 @@
 
 package com.github.gvolpe.fs2rabbit.config
 
-case class Fs2RabbitConfig(
+import cats.data.NonEmptyList
+
+case class Fs2RabbitNodeConfig(
     host: String,
-    port: Int,
+    port: Int
+)
+
+case class Fs2RabbitConfig(
+    nodes: NonEmptyList[Fs2RabbitNodeConfig],
     virtualHost: String,
     connectionTimeout: Int,
     ssl: Boolean,
     username: Option[String],
     password: Option[String],
     requeueOnNack: Boolean,
-    internalQueueSize: Option[Int]
+    internalQueueSize: Option[Int],
+    automaticRecovery: Boolean
 )
+
+object Fs2RabbitConfig {
+  def apply(
+      host: String,
+      port: Int,
+      virtualHost: String,
+      connectionTimeout: Int,
+      ssl: Boolean,
+      username: Option[String],
+      password: Option[String],
+      requeueOnNack: Boolean,
+      internalQueueSize: Option[Int],
+      automaticRecovery: Boolean = true
+  ): Fs2RabbitConfig = Fs2RabbitConfig(
+    nodes = NonEmptyList.one(Fs2RabbitNodeConfig(host, port)),
+    virtualHost = virtualHost,
+    connectionTimeout = connectionTimeout,
+    ssl = ssl,
+    username = username,
+    password = password,
+    requeueOnNack = requeueOnNack,
+    internalQueueSize = internalQueueSize,
+    automaticRecovery = automaticRecovery
+  )
+}
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/ConnectionStream.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/interpreter/ConnectionStream.scala
@@ -17,7 +17,7 @@
 package com.github.gvolpe.fs2rabbit.interpreter
 
 import javax.net.ssl.SSLContext
-
+import cats.data.NonEmptyList
 import cats.effect.Sync
 import cats.syntax.apply._
 import cats.syntax.flatMap._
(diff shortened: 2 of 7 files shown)


ksonj:
looks good to me :+1:

LMnet:
Recently I updated fs2-rabbit to the 1.1.0 version with this change. I'm using pureconfig to read the `Fs2RabbitConfig`. I saw in the release notes that I need to add pureconfig-cats to make the new config works. I did it, and it's not enough. 

I had config like this:
```
rabbit {
  host = "127.0.0.1"
  port = 35672
  username = "guest"
  password = "guest"
  virtual-host = "/"
  connection-timeout = 10000
  ssl = false
  requeue-on-nack = false
}
```

And now I need to have config like this: 

```
rabbit {
  nodes = { 
    head = {
      host = "127.0.0.1"
      port = 35672
    }
    tail = []
  }
  username = "guest"
  password = "guest"
  virtual-host = "/"
  connection-timeout = 10000
  ssl = false
  requeue-on-nack = false
  automatic-recovery = true
}
```

And this config structure is definitely _not_ what I want to see in my config file. Also, I need to add extra `automatic-recovery` flag.

If we want to make fs2-rabbit friendly to use with pureconfig, I believe that the current situation with `Fs2RabbitConfig` should be changed. I can suggest 2 options:
1. Don't change `Fs2RabbitConfig`, but provide an extra `pureconfig-fs2-rabbit` package with some custom `ConfigConvert` instances.
2. Change `Fs2RabbitConfig`. In particular, use `List` instead of `NonEmptyList`.

I prefer the first option because usage of `List` instead of `NonEmptyList` means losing some type-safety.

JanGe:
@LMnet Are you missing the the import for the `pureconfig-cats` module?
```scala
private val config = {
    import pureconfig.generic.auto._
    import pureconfig.module.cats._
    pureconfig.loadConfigOrThrow[Config]
}
```

JanGe:
Regarding the `automatic-recovery` field or other properties with a reasonable default value, I'd suggest [providing a `reference.conf`](https://github.com/lightbend/config#longer-examples) instead.

LMnet:
@JanGe No, I'm not missing `pureconfig-cats` import. My second example uses `ConfigConvert` from this module. And the first example didn't work at all, because pureconfig can't use custom apply method.

LMnet:
I can suggest one more option: fix `pureconfig-cats` to be able to read `NonEmptyList` from the simple array. And actually, I think it would be the best option.

gvolpe:
@LMnet thanks for raising the issue. I'd really like to keep the compatibility with `pureconfig` and I think providing a `pureconfig-fs2rabbit` module would probably make sense. I'm going on holidays this evening but I'll definitely take some time to look into this, get it fixed and publish a new version when I'm back (the weekend of 2nd / 3rd).

PR #158, opened 2019-01-18
Provide an option to allocate connection using cats.effect.Resource

iRevive:
Hi @gvolpe ,

from my perspective, `cats.effect.Resource` can be used to manage the AMQP connection. This change provides more flexible API. 

I have one open question about my PR: personally, I don't like names of methods in my pull request. `createConnectionChannelResource` doesn't look good for me.

Probably, it wiill be better to expose connectionBuilder as a part of public Fs2Rabbit API.
Then naming will be much more clear and obvious:
```scala
for {
  rabbit <- Fs2Rabbit(config)
  _ <- rabbit.connectionBuilder.resource.use(channel => effectProgram.apply(channel)) 
 // or
  _ <- ResilientStream.run(rabbit.connectionBuilder.stream.flatMap(channel => streamProgram.apply(channel)))
} yield ExitCode.Success
```

What do you think?

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Connection.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Connection.scala
@@ -16,8 +16,9 @@
 
 package com.github.gvolpe.fs2rabbit.algebra
 
+import cats.effect.Resource
 import com.github.gvolpe.fs2rabbit.model.AMQPChannel
 
 trait Connection[F[_]] {
-  def createConnectionChannel: F[AMQPChannel]
+  def createConnectionChannel: Resource[F, AMQPChannel]
 }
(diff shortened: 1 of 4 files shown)


ksonj:
Sorry, I was quite caught up in work stuff in the new year. I still intend to bring #150 forward, though, and am hoping to allocate some time over the next week. The changes introduced by this PR are already included (without changing the algebra as suggested by @gvolpe, minus the change of name to `ConnectionBuilder`). While I am not at all opposed to splitting things up into smaller PRs, I think the bulk of the work has already been done. I basically only need to switch back to `Stream` for the consumer (as discussed in #150) and then the examples and docs need some updating.

That being said, I don't see why we could not merge this one first.

gvolpe:
Thanks for your feedback @ksonj !

Given that #150 will be continued I'll close this PR since there's no urgency to have this merged I believe. Anyway, thanks for the proposal @iRevive and feel free to jump in onto #150 if you have any other ideas :smiley:

PR #160, opened 2019-01-19
Remove match clause for scala version 2.10

yoohaemin:
Small cleanup, no need to match 2.10 anymore


yoohaemin:
Would you like a separate PR for that?

PR #168, opened 2019-01-28
Spinning a single Docker instance per suite + polishing tests

gvolpe:
Previously, a `docker` instance was being started per single test. I just created a suite so we only spin up a single instance to run tests faster.

There were also some side-effectful methods in tests (`randomQueueData()` and `randomString()`) that I don't know when they made it into master but I just refactored them.

@iRevive wanna have a look?


iRevive:
Reasonable point regarding `randomQueueData()`. Impure data generation breaks referential transparency for sure. I will keep it in mind, thanks!

PR #171, opened 2019-02-06
Support of RPC pattern

iRevive:
Hi @gvolpe. 

I added a new method `declareQueue` which doesn't take arguments. It creates a queue with a random name. This functionality is necessary for RPC pattern.

This PR should cover https://github.com/gvolpe/fs2-rabbit/issues/114.

### DIFF
--- a/examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
+++ b/examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2017-2019 Gabriel Volpe
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.gvolpe.fs2rabbit.examples
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.UUID
+
+import cats.data.{Kleisli, NonEmptyList}
+import cats.effect._
+import cats.syntax.applicative._
+import cats.syntax.functor._
+import com.github.gvolpe.fs2rabbit.config.declaration.DeclarationQueueConfig
+import com.github.gvolpe.fs2rabbit.config.{Fs2RabbitConfig, Fs2RabbitNodeConfig}
+import com.github.gvolpe.fs2rabbit.effects.MessageEncoder
+import com.github.gvolpe.fs2rabbit.interpreter.Fs2Rabbit
+import com.github.gvolpe.fs2rabbit.model._
+import fs2.Stream
+
+object RPCDemo extends IOApp {
+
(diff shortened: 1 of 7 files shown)


examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+}
+
+class RPCClient[F[_]: Sync](rpcQueue: QueueName)(implicit R: Fs2Rabbit[F], channel: AMQPChannel) {
+
+  private val EmptyExchange = ExchangeName("")
+
+  implicit val encoder: MessageEncoder[F, AmqpMessage[String]] =
+    Kleisli[F, AmqpMessage[String], AmqpMessage[Array[Byte]]](s => s.copy(payload = s.payload.getBytes(UTF_8)).pure[F])
```
gvolpe: Have you checked whether `Kleisli(s => s.copy(payload = s.payload.getBytes(UTF_8)).pure[F])` works here or not? Sometimes it is necessary to help `scalac` with types since it is not smart enough but since the type is already specified for the `encoder` variable it might just work?

examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
```diff
@@ -0,0 +1,128 @@
...
+class RPCServer[F[_]: Sync](rpcQueue: QueueName)(implicit R: Fs2Rabbit[F], channel: AMQPChannel) {
+
+  private val EmptyExchange = ExchangeName("")
+
+  private implicit val encoder: MessageEncoder[F, AmqpMessage[String]] =
+    Kleisli[F, AmqpMessage[String], AmqpMessage[Array[Byte]]](s => s.copy(payload = s.payload.getBytes(UTF_8)).pure[F])
+
+  def serve: Stream[F, Unit] =
+    for {
+      _        <- R.declareQueue(DeclarationQueueConfig.default(rpcQueue))
```
gvolpe: How about this?

```scala
def serve: Stream[F, Unit] =
  R.declareQueue(DeclarationQueueConfig.default(rpcQueue)) >>
    R.createAutoAckConsumer(rpcQueue).flatten.flatMap(handler).drain
```

Compiles in my head :laughing:

examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
iRevive: This example works fine. Also, IntelliJ IDEA can resolve types as well.

examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
iRevive: Looks much better, thanks.

gvolpe:
@iRevive good to merge then?

iRevive:
Sure

examples/src/main/scala/com/github/gvolpe/fs2rabbit/examples/RPCDemo.scala
gvolpe: I meant that you don't need to explicitly add the type like `Kleisli[F, A, B]`, not sure if I made myself very clear there.

iRevive: @gvolpe oh, I get it. It works without explicit parameters too 
`Kleisli(s => s.copy(payload = s.payload.getBytes(UTF_8)).pure[F])`

gvolpe: Cool no worries, I'll do some polishing before publishing a new release (probably this weekend) :)

PR #187, opened 2019-04-25
Allow multiple channels to be created for a single connection

DougC:
[https://github.com/gvolpe/fs2-rabbit/issues/186]

The aim here is to allow a client to create a connection out of a ConnectionStream and then use that to create multiple channels that different parts of the system can use. This supports the model recommended for rabbitmq where different threads operate on separate channels.

* Added a `createConnection` method to the `Connection[F[_]]` trait alongside the existing `createConnectionChannel`
* Created a `Channeller` trait (named to avoid clash with rabbit `Channel` class, up for a better name) with a `createChannel` method
* Created a `ChannelStream` implementation that will create a stream of channels out of a provided connection
* Created an `AMQPConnection` model to wrap a rabbit connection, in line with the `AMQPChannel` model
* Bumped version to 1.2.1-SNAPSHOT for the moment

### DIFF
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Channeller.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Channeller.scala
@@ -18,7 +18,6 @@ package com.github.gvolpe.fs2rabbit.algebra
 
 import com.github.gvolpe.fs2rabbit.model.AMQPChannel
 
-// TODO: Find a better name - I've called it this to avoid clash with the rabbit class called Channel
 trait Channeller[F[_]] {
   def createChannel: F[AMQPChannel]
 }
--- a/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Channeller.scala
+++ b/core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Channeller.scala
@@ -16,10 +16,9 @@
 
 package com.github.gvolpe.fs2rabbit.algebra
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/github/gvolpe/fs2rabbit/algebra/Channeller.scala
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.gvolpe.fs2rabbit.algebra
+
+import com.github.gvolpe.fs2rabbit.model.AMQPChannel
+
+// TODO: Find a better name - I've called it this to avoid clash with the rabbit class called Channel
```
gvolpe: Naming is hard :smile: 

It's fine and not so important in this case since it's just an internal algebra so we can leave it as is. Please remove the TODO comment :)

DougC: Okay, I've removed the TODO. I haven't done any unit tests specifically for this. Is that okay?

gvolpe: Tests are always nice to have but it can be done in a different PR.

What I'm mostly concerned about is on whether we might run into thread-safety issues when creating multiple channels from the same connection. It would be great to add some tests for that (all tests run against a dockerized RabbitMQ).

PR #213, opened 2019-06-27
Add in missing AmqpHeaderVal classes

### LINKED ISSUE #211: `AmqpHeaderVal` features incomplete listing of types
In particular currently, `AmqpHeaderVal` only has support for `Int`, `Long`, `String`, `Integer`, and `java.util.List[_]`. However, the underlying Java type being converted could be any one of the types here: https://github.com/rabbitmq/rabbitmq-java-client/blob/b55bd20a1a236fc2d1ea9369b579770fa0237615/src/main/java/com/rabbitmq/client/impl/ValueWriter.java#L128 

If one of those Java types is found, the library blows up with a `MatchError` that looks like the following:

```scala
scala.MatchError: true (of class java.lang.Boolean)
        at dev.profunktor.fs2rabbit.model$AmqpHeaderVal$.from(model.scala:99)
        at dev.profunktor.fs2rabbit.model$AmqpProperties$.$anonfun$from$5(model.scala:144)
        at scala.collection.StrictOptimizedMapOps.map(StrictOptimizedMapOps.scala:30)
        at scala.collection.StrictOptimizedMapOps.map$(StrictOptimizedMapOps.scala:29)
        at scala.collection.immutable.HashMap.map(HashMap.scala:39)
        at dev.profunktor.fs2rabbit.model$AmqpProperties$.from(model.scala:143)
        at dev.profunktor.fs2rabbit.interpreter.AmqpClientEffect$$anon$1.handleDelivery(AMQPClientEffect.scala:59)
        at com.rabbitmq.client.impl.ConsumerDispatcher$5.run(ConsumerDispatcher.java:149)
        at com.rabbitmq.client.impl.ConsumerWorkService$WorkPoolRunnable.run(ConsumerWorkService.java:104)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

The library should be able to support any of those types in that Java file, otherwise it will blow up at runtime on decoding certain message headers.

changlinli:
These classes need to reflect what's being generated by `com.rabbitmq.client.impl.ValueWriter` otherwise we'll end up with runtime MatchError exceptions.

Closes #211


changlinli:
Note that I don't quite want to merge this yet, I've got some tests I'm writing to validate that #211 is quashed.

Also this is a breaking change since `AmqpHeaderVal` is public.

Note I've also made `ArrayVal` immutable.

changlinli:
Now the tests are in. I'll need to brush up some of the comments that end half-way through, but I'd love for you to take a look at this point @gvolpe

core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
```diff
@@ -82,26 +82,70 @@ object model {
...
+    final case class LongVal(value: Long)                        extends AmqpHeaderVal
+    final case class StringVal(value: String)                    extends AmqpHeaderVal
+    final case class LongStringVal(value: LongString)            extends AmqpHeaderVal
+    final case class ArrayVal(v: Vector[AmqpHeaderVal])          extends AmqpHeaderVal
+    case object NullVal                                          extends AmqpHeaderVal
+
+    /**
+      * This method is meant purely to translate the output of [[com.rabbitmq.client.impl.ValueReader.readFieldValue]]. As such
+      */
     def from(value: AnyRef): AmqpHeaderVal = value match {
```
gvolpe: Having a fresh look I think we should change the signature of this method to return `Either[HeaderParsingError, AmqpHeaderVal]` instead.

Since this is called from `def from(basicProps: AMQP.BasicProperties): AmqpProperties` we should change that one too to return `Either[SomeError, AmqpProperties]`.  And where this is called in the consumer logic we should enqueue a `Left(error)` in the internal queue (`AmqpClientEffect.scala`).

```scala
// val props       = AmqpProperties.from(properties) remove this
internals.queue.fold(()) { internalQ =>
  val eitherEnvelope = AmqpProperties.from(properties).map { props =>
    AmqpEnvelope(DeliveryTag(tag), body, props, exchange, routingKey, redelivered)
  }
  internalQ.enqueue1(eitherEnvelope).toIO.unsafeRunAsync(_ => ())
}
```

The custom error type should probably extend `NoStackTrace` to be lightweight and compatible with `cats-effect` to be rethrown in `IO`.

Does that make sense?

changlinli: I think so, but it's a bit tricky. Any error coming from `from` is a bug (in fact it might make sense for the method to be package private as a result; I'm also contemplating renaming it `unsafeFrom`). It indicates that either I didn't understand the underlying Java code as well as I should or the underlying Java code has some bug. It's not really something you can handle at runtime (other than lumping it in with the general error stuff I talk about below).

That is, assuming the message is well-formed at the byte level, `from` should never throw.

Well of course you can't assume the message is well-formed at the byte level! But in that case the code in `from` won't be the source of the error (e.g. it won't be a `MatchError`). `ValueReader` will blow up with some sort of other exception (e.g. `EOF` or some other low-level byte read error). But that's not really a `HeaderParsingError` at that point. Any part of the message could be malformed at the byte level.

I think what makes more sense is wrapping the entire message parsing code in `AmqpClientEffect` in a check to enqueue an error if any part of the parse blows up, if that doesn't already exist.

Thoughts?

changlinli: Ah... this changes if you intend `AmqpProperties.from` to be part of the public API (in addition to creating an `AmqpProperties` from scratch) that users use since users could pass in some sort of malformed `BasicProperties`. Is that your intention?

I mean... you could name that one `unsafe*` too ;), since I assume that most of the time if a user already has a `BasicProperties` on hand they're certain it's valid (e.g. because they got it back from the underlying Java library rather than constructing it on their own) and there's other ways they could mess up that `BasicProperties` (e.g. a non-numeric `String` in the `expiration` field) if they were constructing it by hand.

changlinli:
I made a rather invasive cosmetic change which is renaming `AmqpHeaderVal` to keep it closer to the name in the spec and in the underlying RabbitMQ library. I'm happy to change that back.

Also I'm contemplating enforcing the `ShortString` requirement for the header keys, either by turning the whole field into a `TableVal` or making the keys `ShortString`. Thoughts?

changlinli:
Fascinating... CI is blowing up on a difference between how `BigDecimal` is handled in Scala 2.12 vs 2.13.

```scala
Welcome to Scala 2.12.8 (OpenJDK 64-Bit Server VM, Java 1.8.0_212).
Type in expressions for evaluation. Or try :help.

scala> val upperLimit = BigDecimal(Short.MaxValue)
upperLimit: scala.math.BigDecimal = 32767

scala> val x = BigDecimal(new java.math.BigDecimal("6.107777E+136"))
x: scala.math.BigDecimal = 6.107777E+136

scala> x % upperLimit
res0: scala.math.BigDecimal = 13092
```
vs
```scala
Welcome to Scala 2.13.0 (OpenJDK 64-Bit Server VM, Java 1.8.0_212).
Type in expressions for evaluation. Or try :help.

scala> val upperLimit = BigDecimal(Short.MaxValue)
upperLimit: scala.math.BigDecimal = 32767

scala> val x = BigDecimal(new java.math.BigDecimal("6.107777E+136"))
x: scala.math.BigDecimal = 6.107777E+136

scala> x % upperLimit
java.lang.ArithmeticException: Division impossible
  at java.math.BigDecimal.divideToIntegralValue(BigDecimal.java:1857)
  at java.math.BigDecimal.divideAndRemainder(BigDecimal.java:1989)
  at java.math.BigDecimal.remainder(BigDecimal.java:1923)
  at scala.math.BigDecimal.remainder(BigDecimal.scala:525)
  at scala.math.BigDecimal.$percent(BigDecimal.scala:529)
  ... 36 elided
```

changlinli:
Also fun. Discovered in the course of testing a bug in the underlying Java library: https://github.com/rabbitmq/rabbitmq-java-client/issues/617

core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
gvolpe: `unsafeFrom` in both cases sounds good to me. The possible issue I see is that we need to call `AmqpProperties` in the callback functions of the Java consumer (performing side-effects) that returns `Unit`. Even if we wrap this particular action in `IO` we need to `unsafeRun` it so it's equivalent to just throw an exception from there (`handleDelivery` method). 

If we receive an either instead we can enqueue that Left exception in the internal queue and that would be raised as an error in the pure consumer (as I tried to show in the example above).

We could also keep the methods `from` as they are (just rename them `unsafeFrom`) and from the call site we wrap it as `Try(unsafeFrom(...)).toEither` and again enqueue the either value in our internal queue. 

The important thing is that we shouldn't be throwing exceptions from the Java callback. We should get an exception on the consumer so the user is able to handle it.

> this changes if you intend AmqpProperties.from to be part of the public API 

I don't think people are using this. But if they are then they should be aware it's an *unsafe* action so the rename is fine.

gvolpe:
Interesting bugs, having fun @changlinli ? :smile:

core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
changlinli: Yep I'll add the wrapper around errors in the queue since it's probably good practice anyways (a bug affecting a single message should not bring the whole thing crashing down). Although what I meant is what any byte read errors would occur before the consumer callback is even called (because they would surface while trying to construct the `Envelope` and `BasicProperties`). And I don't know how that would affect this library.

changlinli:
Yak shaving is... a mix of emotions...

changlinli:
Ugh... generating Scaladocs fail with 

```
[error] /home/changlin/Projects/fs2-rabbit/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala:114:3: Could not find any member to link for "com.rabbitmq.client.impl.ValueReader.readFieldValue".
[error]   /**
[error]   ^
[error] /home/changlin/Projects/fs2-rabbit/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala:165:5: Could not find any member to link for "BigDecimal".
[error]     /**
[error]     ^
[error] /home/changlin/Projects/fs2-rabbit/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala:139:5: Could not find any member to link for "AmqpFieldValue.unsafeFromValueReaderOutput".
[error]     /**
[error]     ^
[error] /home/changlin/Projects/fs2-rabbit/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala:192:7: Could not find any member to link for "BigDecimal".
[error]       /**
[error]       ^
[error] four errors found
```

Dunno why BigDecimal is bad, but the others are because they're package-private. What should I do about that? I can imagine:

+ Get rid of the `[[...]]`
+ Make them public instead of package private
+ Turn on `-no-link-warnings` as a `scalac` option

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
```diff
@@ -17,25 +17,25 @@
...
 import dev.profunktor.fs2rabbit.model._
-import com.rabbitmq.client._
 
 class AmqpClientEffect[F[_]: Effect] extends AMQPClient[F] {
 
   private[fs2rabbit] def defaultConsumer[A](
       channel: Channel,
       internals: AMQPInternals[F]
-  ): F[Consumer] = Applicative[F].pure {
+  ): F[Consumer] = Effect[F].delay {
```
gvolpe: Use `Sync[F].delay` instead

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
```diff
@@ -52,18 +52,23 @@ class AmqpClientEffect[F[_]: Effect] extends AMQPClient[F] {
...
-        val exchange    = ExchangeName(envelope.getExchange)
-        val redelivered = envelope.isRedeliver
-        val props       = AmqpProperties.from(properties)
-        internals.queue.fold(()) { internalQ =>
-          val envelope = AmqpEnvelope(DeliveryTag(tag), body, props, exchange, routingKey, redelivered)
-          internalQ
-            .enqueue1(Right(envelope))
-            .toIO
-            .unsafeRunAsync(_ => ())
+        val tryEnqueuing = scala.util.Try {
```
gvolpe: Mmm this is not what I had in mind. Only try the `AmqpProperties.unsafeFrom` and convert it `toEither` so you can `map` that into the creation of the envelope and enqueue that (you now have an either). 

```scala
        val tag         = envelope.getDeliveryTag
        val routingKey  = RoutingKey(envelope.getRoutingKey)
        val exchange    = ExchangeName(envelope.getExchange)
        val redelivered = envelope.isRedeliver    
        internals.queue.fold(()) { internalQ =>
          val env = Try(AmqpProperties.unsafeFrom(properties)).toEither.map { props =>
            AmqpEnvelope(DeliveryTag(tag), body, props, exchange, routingKey, redelivered)
          }
          internalQ.enqueue1(env).toIO.unsafeRunAsync(_ => ())
        }
```

core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
```diff
@@ -81,27 +87,212 @@ object model {
...
-    final case class IntVal(value: Int)               extends AmqpHeaderVal
-    final case class LongVal(value: Long)             extends AmqpHeaderVal
-    final case class StringVal(value: String)         extends AmqpHeaderVal
-    final case class ArrayVal(v: collection.Seq[Any]) extends AmqpHeaderVal
+  /**
+    * This hierarchy is meant to reflect the output of
+    * [[com.rabbitmq.client.impl.ValueReader.readFieldValue]] in a type-safe
+    * way.
+    *
+    * Note that we don't include LongString here because of some ambiguity in
```
gvolpe: Love the docs :+1:

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
changlinli: Ah of course... this is what late night coding mixed with over-ambitious refactoring gets you. Sorry about that (I was trying to wrap all the `val`s in a `Try` then somehow decided to just use `fromTry` and then everything went downhill from there).

changlinli: Done with 68df4d9. I still ended up wrapping the whole chunk in a `Try` rather than just the `unsafeFrom`, both because it is (very marginally) more conservative and also because to my eyes it reads better (`unsafeFrom` isn't privileged in any way over the other assignments). Thoughts?

changlinli:
RE Scaladocs, I still like the `[[` because they're a nice convenience in an IDE, where they offer jump to definition and also are automatically renamed when you rename the method itself. I'd be more inclined to turn on `-no-link-warnings` (that's what I've done on other projects).

Thoughts?

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
changlinli: Done with 76c40be

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
gvolpe: No problems :) I think only `unsafeFrom` should be wrapped in the `Try` block (it's very unfortunate `toEither` doesn't work in `2.11.x`...). The other stuff are merely variable bindings with no side-effects.

changlinli:
Oh one last thing is inconsistent naming in this PR. I have `ShortString.of` and `from*` in other places. What kind of naming do you prefer? `of`, `from`, `from[name of type]`?

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
gvolpe: And I still prefer `internals.queue.fold(())` over `internals.queue.fold(Applicative[F].unit)` since we still need to return `Unit` via `unsafeRun`. The extra layer seems unnecessary.

gvolpe:
Let's keep `from` for consistency :)

changlinli:
Also do `sealed abstract case class`es need `private` modifiers to their constructors/would you like them to be there as a reminder? I just realized the `private` might be redundant in all those ones I've written... (but it doesn't harm things)

gvolpe:
That seems fine to me :+1: You can leave it as is for now

changlinli:
One more style question, I just realized that now `TimestampVal.from` is overloaded. Do you prefer overloaded methods or should I suffix `from` with specific types to prevent overloading?

gvolpe:
If it's only two overloaded methods that is fine :+1:

changlinli:
Cool, in that case I think I'm done with another pass at this PR and would love for you to take another look.

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
gvolpe: You might have forgotten this @changlinli :) Other than this it looks ready to merge.

gvolpe:
I just did @changlinli , see my other comment.

changlinli:
Apologies for taking so long to get back to this; I've submitted another commit now.

changlinli:
@gvolpe Sorry about taking so long; I completely forgot to check whether my fix for Scala 2.11 compatibility passed CI...

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
```diff
@@ -71,15 +71,23 @@ class AmqpClientEffect[F[_]: Effect] extends AMQPClient[F] {
...
-        // Manual right projection here to keep compatibility with Scala 2.11
-        val envelopeOrErr = amqpPropertiesOrErr.right
-          .map { props =>
-            val tag         = envelope.getDeliveryTag
-            val routingKey  = RoutingKey(envelope.getRoutingKey)
-            val exchange    = ExchangeName(envelope.getExchange)
-            val redelivered = envelope.isRedeliver
-            AmqpEnvelope(DeliveryTag(tag), body, props, exchange, routingKey, redelivered)
-          }
+        // This manual match instead of a map is because of two annoying things:
```
gvolpe: Doesn't `import cats.implicits._` solve the issue for all the scala versions?

gvolpe:
No worries @changlinli , thanks for putting up the work! :smiley:

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/AMQPClientEffect.scala
changlinli: Unfortunately it shows up as an unused import for Scala 2.12 and Scala 2.13 and bombs out with a fatal warning

changlinli: I could collapse the fine-grained imports right now at the top of the file into a single `cats.implicits._` which would solve it.

changlinli: Or I could invoke the `Functor` instance manually

PR #215, opened 2019-07-01
Add option to pass in SASL config

### LINKED ISSUE #212: Have to hack around to support client side SSL certificate authentication (SASL EXTERNAL)
Although RabbitMQ comes with a plugin for doing SSL-based authentication which is supported by the RabbitMQ Java client library, `fs2-rabbit` doesn't expose that. There is a workaround for it, which looks something like the following:

```scala
package dev.profunktor.fs2rabbit

import cats.effect.ConcurrentEffect
import cats.syntax.functor._
import com.rabbitmq.client.DefaultSaslConfig
import dev.profunktor.fs2rabbit.config.Fs2RabbitConfig
import dev.profunktor.fs2rabbit.interpreter.{AmqpClientEffect, ConnectionEffect, Fs2Rabbit, LiveInternalQueue}
import dev.profunktor.fs2rabbit.program.{AckingProgram, ConsumingProgram}
import javax.net.ssl.SSLContext

object Blah {

  def customCreateFs2Rabbit[F[_] : ConcurrentEffect](config: Fs2RabbitConfig, sslContext: Option[SSLContext]): F[Fs2Rabbit[F]] = {
    ConnectionEffect.mkConnectionFactory[F](config, sslContext).map {
      case (factory, addresses) =>
        val saslConfig = DefaultSaslConfig.EXTERNAL
        factory.setSaslConfig(saslConfig)
        val amqpClient = new AmqpClientEffect[F]
        val conn       = new ConnectionEffect[F](factory, addresses)
        val internalQ  = new LiveInternalQueue[F](config.internalQueueSize.getOrElse(500))
        val acker      = new AckingProgram[F](config, amqpClient)
        val consumer   = new ConsumingProgram[F](amqpClient, internalQ)
        new Fs2Rabbit[F](conn, amqpClient, acker, consumer)
    }
  }
}
```

which you then use to create an `Fs2Rabbit[F]` instead of the usual way, but because `mkConnectionFactory` is `private[fs2rabbit]`, this file has to be put in a `dev.profunktor.fs2rabbit` namespace, which is kind of annoying. Not a show-stopper, but is a bit of a rough edge for the library.

changlinli:
This allows us to support certificate-based SSL authentication.

Fixes #212

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/Fs2Rabbit.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/Fs2Rabbit.scala
@@ -33,6 +33,8 @@ object Fs2Rabbit {
   def apply[F[_]: ConcurrentEffect](
       config: Fs2RabbitConfig,
       sslContext: Option[SSLContext] = None,
+      // Unlike SSLContext, SaslConfig is not optional because it is always set
+      // by the underlying Java library, even if the user doesn't set it.
       saslConfig: SaslConfig = DefaultSaslConfig.PLAIN
   ): F[Fs2Rabbit[F]] =
     ConnectionEffect.mkConnectionFactory[F](config, sslContext, saslConfig).map {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/Fs2Rabbit.scala
```diff
@@ -31,9 +32,12 @@ import javax.net.ssl.SSLContext
 object Fs2Rabbit {
   def apply[F[_]: ConcurrentEffect](
       config: Fs2RabbitConfig,
-      sslContext: Option[SSLContext] = None
+      sslContext: Option[SSLContext] = None,
+      // Unlike SSLContext, SaslConfig is not optional because it is always set
+      // by the underlying Java library, even if the user doesn't set it.
```
gvolpe: Interesting. TIL.

Issue #230, opened 2019-08-29
Separate thread pool for blocking publish operations


alexander-branevskiy:
Seems that the current implementation of message publishing has some deficiencies. It uses native Rabbit MQ client as a transport and its ```basicPublish``` operation can block a thread what can lead to several problems like API freezing (in case if it uses the same thread pool), overall performance degradation, etc... So i suggest either moving all operations related to ```basicPublish``` to a separate thread pool (fixed?) or try to use NIO rabbit MQ client API (not sure whether it will help) in order to resolve the problem.

gvolpe:
In the meantime, I think you can get away by wrapping the publishing call in a blocking execution context. Either using `ContextShift` or `Blocker` if using a newer version of `cats-effect`. Something like:

```scala
def publishSimpleMessage[F[_]: Sync](blocker: Blocker[F], publisher: String => F[Unit]): F[Unit] =
  blocker.blockOn(publisher("Hello world!"))
```

Or in older versions:

```scala
def publishSimpleMessage[F[_]: ContextShift: Sync](publisher: String => F[Unit]): F[Unit] =
  ContextShift[F].evalOn(publisher("Hello world!"))
```

PR #241, opened 2019-09-15
Preparing release 2.0.0

gvolpe:
Closes #234 and closes #235 

- Dropping support for Scala 2.11.x as Circe does not support it. 
- Updating and unifying dependencies for the different Scala versions. 
- Bumping version.

Pinging @Daenyth to get the green light on this one.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ name := """fs2-rabbit-root"""
 organization in ThisBuild := "dev.profunktor"
 
 scalaVersion in ThisBuild := "2.12.9"
-crossScalaVersions in ThisBuild := Seq("2.12.9", "2.13.0")
+crossScalaVersions in ThisBuild := Seq("2.12.10", "2.13.0")
 
 sonatypeProfileName := "dev.profunktor"
 
--- a/build.sbt
+++ b/build.sbt
@@ -8,28 +8,20 @@ name := """fs2-rabbit-root"""
 organization in ThisBuild := "dev.profunktor"
 
 scalaVersion in ThisBuild := "2.12.9"
-crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.9", "2.13.0")
-
-// makes `tut` fail :( -> https://github.com/tpolecat/tut/issues/255
-//scalaVersion in ThisBuild := "2.12.8" // needed for metals
+crossScalaVersions in ThisBuild := Seq("2.12.9", "2.13.0")
 
 sonatypeProfileName := "dev.profunktor"
 
-promptTheme := PromptTheme(List(
-  text("[sbt] ", fg(105)),
-  text(_ => "fs2-rabbit", fg(15)).padRight(" λ ")
- ))
+promptTheme := PromptTheme(
+  List(
+    text("[sbt] ", fg(105)),
+    text(_ => "fs2-rabbit", fg(15)).padRight(" λ ")
+  )
+)
 
-// We use String as our input type because `scalaVersion.value` cannot be called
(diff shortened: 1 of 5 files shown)


gvolpe:
Leaving this here before I forget... It would be nice to get #230 done before a release. I think I can work on it this week.

ybasket:
Nice - looking forward to this release!

Maybe upgrade to Scala 2.12.10 as well?

gvolpe:
Yeah, I can do that too if there's no big issues! I didn't know 2.12.10 was out :smile:

build.sbt
```diff
@@ -8,28 +8,20 @@ name := """fs2-rabbit-root"""
 organization in ThisBuild := "dev.profunktor"
 
 scalaVersion in ThisBuild := "2.12.9"
-crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.9", "2.13.0")
-
-// makes `tut` fail :( -> https://github.com/tpolecat/tut/issues/255
-//scalaVersion in ThisBuild := "2.12.8" // needed for metals
+crossScalaVersions in ThisBuild := Seq("2.12.9", "2.13.0")
```
Daenyth: Do we want to set this to 2.12.10?
```suggestion
crossScalaVersions in ThisBuild := Seq("2.12.10", "2.13.0")
```

build.sbt
```diff
@@ -117,44 +91,51 @@ lazy val noPublish = Seq(
...
+lazy val `fs2-rabbit-circe` = project
+  .in(file("json-circe"))
   .settings(commonSettings: _*)
   .settings(libraryDependencies ++= JsonDependencies(scalaVersion.value))
   .settings(parallelExecution in Test := false)
   .enablePlugins(AutomateHeaderPlugin)
   .dependsOn(`fs2-rabbit`)
 
-lazy val `fs2-rabbit-test-support` = project.in(file("test-support"))
+lazy val `fs2-rabbit-test-support` = project
```
Daenyth: Is this published? I should check that out, might be handy in my project :)

project/Dependencies.scala
```diff
@@ -2,12 +2,15 @@ import sbt._
...
+  object Version {
     val catsEffect = "2.0.0"
     val fs2        = "2.0.0"
-    val circe      = "0.11.1"
+    val circe      = "0.12.1"
     val amqpClient = "5.7.3"
     val logback    = "1.2.3"
+    val monix      = "3.0.0"
+    val zio        = "1.0.0-RC12-1"
+    val zioCats    = "2.0.0.0-RC3"
```
Daenyth: I feel like we should only release ourselves as `RC` if one of our dependencies is at `RC`

version.sbt
```diff
@@ -1,2 +1,2 @@
-version in ThisBuild := "2.0.0-RC3"
+version in ThisBuild := "2.0.0"
```
Daenyth:
```suggestion
version in ThisBuild := "2.0.0-RC4"
```

Because of the zio dependency only. Your call

build.sbt
gvolpe: Yup, I just made the change in my local branch! :smile: I was making sure everything compiles before pushing.

build.sbt
gvolpe: It surely is :) https://search.maven.org/search?q=a:fs2-rabbit-test-support_2.13

project/Dependencies.scala
gvolpe: These are only for the examples.

version.sbt
gvolpe: `monix` and `zio` are only pulled by the `examples` which are not published as a dep.

PR #244, opened 2019-09-17
Creating an Fs2Rabbit client requires a Blocker

### LINKED ISSUE #230: Separate thread pool for blocking publish operations

gvolpe:
closes #230 

@alexander-branevskiy have a look if you can. 
@Daenyth your review would be much appreciated as well.


core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Publishing.scala
```diff
@@ -16,46 +16,53 @@
...
-import com.rabbitmq.client.Channel
 
 trait Publishing[F[_]] {
 
   def createPublisher[A](
       channel: Channel,
       exchangeName: ExchangeName,
-      routingKey: RoutingKey
+      routingKey: RoutingKey,
+      blocker: Blocker
```
Daenyth: I don't believe this belongs on here - it's an implementation detail. Whether `createPublisher` blocks a thread or not depends on how it's implemented. I'd put this param on the implementation constructor of the class that needs to invoke it, and bubble it through the stack from there

examples/src/main/scala/dev/profunktor/fs2rabbit/examples/RPCDemo.scala
```diff
@@ -83,20 +88,25 @@ class RPCClient[F[_]: Sync](rpcQueue: QueueName)(implicit R: Fs2Rabbit[F], chann
...
+      consumer <- Stream.eval(R.createAutoAckConsumer(queue))
+      response <- consumer.filter(_.properties.correlationId.contains(correlationId)).take(1)
+      _        <- Stream.eval(putStrLn(s"[Client] Request $body. Received response [${response.payload}]"))
     } yield response
   }
 
 }
 
-class RPCServer[F[_]: Sync: LiftIO](rpcQueue: QueueName)(implicit R: Fs2Rabbit[F], channel: AMQPChannel) {
+class RPCServer[F[_]: Sync: LiftIO](rpcQueue: QueueName, blocker: Blocker)(
```
Daenyth: `Sync: LiftIO` is strange - is this for the `IO.fromEither`? You can get that via `MonadError[F, Throwable]`, `either.liftTo[F]`

gvolpe: Mmmm I don't know whether it's needed here. It might be a left over.

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Publishing.scala
gvolpe: You mean adding it only in the interpreter but leaving the interface clean? That makes sense.

gvolpe: Also, I'm just realizing that I'm not using the `Blocker`s at all in the implementations :man_facepalming:

gvolpe: So the interpreter `AMQPClientEffect` is the only one needing the `Blocker`. This means that we need to provide a `Blocker` when creating the `Fs2Rabbit` client, even when we don't need to create a publisher. Not sure whether that's the best choice (probably yes). What do you think?

Daenyth: I think it's an improvement over today. I also noticed that constructing Fs2Rabbit requires configuration values that are used either only for publishing or only for consuming.

I'd prefer to tackle splitting up producer/consumer dependencies separately

gvolpe: I went with adding `Blocker` directly when creating an `Fs2Rabbit` client as it makes the usage much easier.

gvolpe:
Ready for review (and copying this message so it doesn't get lost):

> I went with adding Blocker directly when creating an Fs2Rabbit client. Even if you don't need publishing, you need to create a Blocker which might not be ideal but it makes the API usage much easier.

> I don't know, feels like a reasonable choice to make for now but I'd be happy to change this in a future release if users are not happy with it.

core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/Fs2Rabbit.scala
```diff
@@ -30,16 +30,17 @@ import fs2.Stream
 import javax.net.ssl.SSLContext
 
 object Fs2Rabbit {
-  def apply[F[_]: ConcurrentEffect](
+  def apply[F[_]: ConcurrentEffect: ContextShift](
```
Daenyth: 👍 This feels like the most appropriate way to do this.

site/src/main/tut/publishers/index.md
```diff
@@ -6,6 +6,8 @@ number: 10
 
 # Publishers
 
+Publishing are blocking actions in the underlying Java client so you need to pass in a `cats.effect.Blocker` when creating the `Fs2Rabbit` client.
```
Daenyth:
```suggestion
Publishing are blocking actions in the underlying Java client so you need to pass in a [`cats.effect.Blocker`](https://typelevel.org/cats-effect/concurrency/basics.html#choosing-thread-pool) when creating the `Fs2Rabbit` client.
```

PR #285, opened 2019-12-03
Adding support for timestamp property

drobert:
Noticed there wasn't support for the amqp property `timestamp`. Adding it in.

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
@@ -305,6 +305,7 @@ object model {
       expiration: Option[String] = None,
       replyTo: Option[String] = None,
       clusterId: Option[String] = None,
+      timestamp: Option[Date] = None,
       headers: Map[String, AmqpFieldValue] = Map.empty
   )
 
@@ -335,6 +336,7 @@ object model {
         expiration = Option(basicProps.getExpiration),
         replyTo = Option(basicProps.getReplyTo),
         clusterId = Option(basicProps.getClusterId),
+        timestamp = Option(basicProps.getTimestamp),
         headers = Option(basicProps.getHeaders)
           .fold(Map.empty[String, Object])(_.asScala.toMap)
           .map {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
```diff
@@ -305,6 +305,7 @@ object model {
       expiration: Option[String] = None,
       replyTo: Option[String] = None,
       clusterId: Option[String] = None,
+      timestamp: Option[Date] = None,
```
Daenyth: 🔧 `j.util.Date` is super deprecated. This should be `j.time.Instant` instead. I see that the java rabbit library takes `Date` but that seems like a historic artifact rather than good design.

I'd provide the user-friendly api here and eat the conversion from `Instant => Date` internally.

Daenyth:
@gvolpe Note that this will produce a not-binary-compatible update because adding fields to a case class modifies the `copy` method signature.

We should consider changing AmqpProperties to be a trait/abstract class in the future to avoid this kind of breakage.

gvolpe:
@Daenyth thanks for looking into it.

I never planned to support binary breakages to be honest. I consider this library a final user lib so I didn't even bothered setting up MiMa, that's a burden I'm not willing to maintain. 
I have more than enough with the burden of cutting new releases and reviewing some PRs, no other motivation on my side since I no longer use it.

If you would like to take care of this I'd happily grant you all the necessary permissions :)

Daenyth:
I don't have experience with MiMa or care to set it up right now, just pointing it out so we can put it in release notes for the most part, and pick an appropriate version number

Daenyth:
@drobert it looks like some unit tests are failing.

@gvolpe if you add me to the circleci permissions I can also set up test result reporting to make this easier.

core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
```diff
@@ -256,6 +256,7 @@ object model {
       // only read a maximum of 4 bytes before bailing out (similarly it will
       // read no more than the first 8 bits to determine scale).
       case bd: java.math.BigDecimal => DecimalVal.unsafeFrom(bd)
+      case ts: java.time.Instant    => TimestampVal.from(ts)
```
drobert: I'm actually unclear on whether this is safe. It seems reasonable to treat Date and Instant the same way, but it's also called recursively from List and Map types, and I don't know if the 'only 1 second granularity' policy is applicable all the way down.

core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
drobert: Modified in https://github.com/profunktor/fs2-rabbit/pull/285/commits/69563855c62008c544e2198be0e5182a85eddf12

gvolpe:
I'll have a look later and see how I can add you @Daenyth , apparently you need to belong the organization first.

Issue #300, opened 2020-01-01
Acker function does not support `basicReject`


DougC:
The Acker only allows for `Ack` and `NAck`, but not `Reject`

https://github.com/profunktor/fs2-rabbit/blob/f5012a889ff5468e2d69c7cbe2a1cbbdf2793ca1/core/src/main/scala/dev/profunktor/fs2rabbit/program/AckingProgram.scala#L40

DougC:
See https://www.rabbitmq.com/api-guide.html#recovery-and-acknowledgements for discussion of `basicReject`. Reject is useful for supporting deadletter exchanges

gvolpe:
This should be fairly easy to add :wink:

DougC:
PR here https://github.com/profunktor/fs2-rabbit/pull/301 :-)

PR #301, opened 2020-01-05
[issue 300] Support sending a `Reject` result in response to a message

DougC:
Closes https://github.com/profunktor/fs2-rabbit/issues/300

* Added a `Reject` case to `AckResult`.
* Added `basicReject` to `Consume` and the necessary implementations to the other traits and classes.
* Added a `requeueOnReject` boolean parameter to `Fs2RabbitConfig`
* Added tests for requeue and non-requeue on reject

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Consume.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Consume.scala
@@ -50,10 +50,12 @@ object Consume {
                 .unsafeRunAsync(_ => ())
             }
 
-          override def handleDelivery(consumerTag: String,
-                                      envelope: Envelope,
-                                      properties: AMQP.BasicProperties,
-                                      body: Array[Byte]): Unit = {
+          override def handleDelivery(
+              consumerTag: String,
+              envelope: Envelope,
+              properties: AMQP.BasicProperties,
+              body: Array[Byte]
+          ): Unit = {
             def rewrappedError(err: Throwable) =
               Left(
                 new Exception(
@@ -67,7 +69,8 @@ object Consume {
                 $properties
                 """,
                   err
-                ))
+                )
+              )
 
             val amqpPropertiesOrErr =
               Try(AmqpProperties.unsafeFrom(properties)) match {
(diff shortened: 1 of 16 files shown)


DougC:
> Looks good to me, only need to update the docs to pass the CI build :)

Ah, okay. Where do I do that? The microsite? It wasn't at all clear to me why the CI build was failing

DougC:
Found it!

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Consume.scala
```diff
@@ -114,6 +117,10 @@ object Consume {
           channel.value.basicNack(tag.value, multiple, requeue)
         }
 
+      override def basicReject(channel: AMQPChannel, tag: DeliveryTag, requeue: Boolean): F[Unit] = Sync[F].delay {
+        channel.value.basicReject(tag.value, requeue)
```
Daenyth: Does this need `Blocker`?

tests/src/test/scala/dev/profunktor/fs2rabbit/interpreter/Fs2RabbitSpec.scala
```diff
@@ -477,6 +503,32 @@ trait Fs2RabbitSpec { self: BaseSpec =>
       }
   }
 
+  it should "requeue a message in case of Reject when option 'requeueOnReject = true'" in withStreamRejectRabbit {
```
Daenyth: I'd kill this test, it's trivially true if your first test is true because of the monad laws with regard to `Stream.eval`

DougC: I can do, but I was replicating the same test as for the `NAck` case. But also, it confirms that the flag has the right effect.

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Consume.scala
DougC: `Consume` could be implemented with `Blocker`, but it isn't currently and I added the `basicReject` method based on the implementation of the other methods in the object. I was aiming to minimise the amount of change in the PR, so maybe using `Blocker` vs `Sync` is something that should be should be reviewed across the codebase as a separate issue?

Daenyth: > using Blocker vs Sync is something that should be should be reviewed across the codebase as a separate issue

👍 very reasonable

tests/src/test/scala/dev/profunktor/fs2rabbit/interpreter/Fs2RabbitSpec.scala
Daenyth: @gvolpe thoughts?

gvolpe: I think it's fine. Improving the test suite can be another PR, to be fair.

Issue #314, opened 2020-02-07
Add cats instances for `AmqpEnvelope`


Daenyth:
I've found it useful to have `Functor[AmqpEnvelope]` and the implementation is trivial, so we should add that and anything else useful (Eq/Show?)

```scala

  implicit val envelopeInstance: Functor[AmqpEnvelope] =
    new Functor[AmqpEnvelope] {

      override def map[A, B](fa: AmqpEnvelope[A])(f: A => B): AmqpEnvelope[B] =
        fa.copy[B](payload = f(fa.payload))
    }
```

gvolpe:
Fixed in #364

PR #364, opened 2020-06-08
Cats instances

bpholt:
This PR addresses #314 by adding a `Traverse[AmqpEnvelope]`, `CommutativeSemigroup[DeliveryTag]`, and `Order` or `Eq` instances for most of the types in `dev.profunktor.fs2rabbit.model`. The lawfulness of all the instances is tested using Discipline.

I also exposed the ScalaCheck `Arbitrary` and `Cogen` instances used to test the cats instances in a new `testkit` module. I originally wrote several of the `Arbitrary` instances for property testing in a downstream app, and it would be nice to be able to reuse them in both places.

`Order` instances are provided where the underlying types allow, with `Eq` instances otherwise. In particular, the scodec, `Map`, and `Array`-based types are `Eq`-only.

The idea behind the `CommutativeSemigroup[DeliveryTag]` instance is to use it for acknowledging multiple messages. The instance takes the max of the `DeliveryTag`'s underlying values.

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
@@ -22,15 +22,17 @@ import java.time.Instant
 import java.time.temporal.ChronoUnit
 import java.util.Date
 
-import cats.data.Kleisli
+import cats.data._
 import cats.implicits._
-import cats.{Applicative, ApplicativeError}
+import cats.kernel.CommutativeSemigroup
+import cats._
 import com.rabbitmq.client.{AMQP, Channel, Connection, LongString}
 import dev.profunktor.fs2rabbit.arguments.Arguments
 import dev.profunktor.fs2rabbit.effects.{EnvelopeDecoder, MessageEncoder}
 import dev.profunktor.fs2rabbit.javaConversion._
 import fs2.Stream
 import scodec.bits.ByteVector
+import scodec.interop.cats._
 
 object model {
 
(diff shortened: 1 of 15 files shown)


core/src/main/scala/dev/profunktor/fs2rabbit/model.scala
```diff
@@ -388,6 +494,31 @@ object model {
...
+
+      override def foldLeft[A, B](fa: AmqpEnvelope[A], b: B)(f: (B, A) => B): B =
+        f(b, fa.payload)
+
+      override def foldRight[A, B](fa: AmqpEnvelope[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+        f(fa.payload, lb)
+    }
+
+    implicit def eqAmqpEnvelope[A](implicit A: Eq[A]): Eq[AmqpEnvelope[A]] =
+      Eq.and(
```
Daenyth, line 509: ⛏️ Not really a fan of this approach given that these are case classes; it seems perfectly reasonable to use the normal Eq.fromUniversal

gvolpe: I don't hold a strong opinion on this one. Let us know what you think @bpholt, if there are no further changes we can merge it :)

bpholt: I went back and forth on that too, but I think this approach works better because of the headers in the `properties: AmqpProperties` field, which I believe ends up benefiting from the `Eq[AmqpFieldValue]` instance for the values in the map. (It may work better for certain payload types too?)

PR #368, opened 2020-06-10
include fs2-rabbit-testkit in root aggregate so that it gets published

bpholt:


gvolpe:
There you go `0.0.0+1-8778c704-SNAPSHOT` :)

bpholt:
Thanks!

PR #388, opened 2020-08-16
Add optional `MetricsCollector` to the client

kubum:
Motivation:

We would like to client statistics. Luckily, RabbitMQ Java Client includes implementations of Dropwizard and Micrometer metrics collectors. Let's expose them.

Modification:

- Add `metricsCollector` to the resource constructors.
- Add an example application that exposes metrics via JMX.
- Add an example into the microsite documentation.

Result:

- See JMX metrics via JConsole with Java Client 5.9.0 and Dropwizard 4.1.5

image

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
@@ -19,7 +19,7 @@ package dev.profunktor.fs2rabbit.algebra
 import cats.data.NonEmptyList
 import cats.effect.{Resource, Sync}
 import cats.implicits._
-import com.rabbitmq.client.{Address, ConnectionFactory, DefaultSaslConfig, SaslConfig}
+import com.rabbitmq.client.{Address, ConnectionFactory, DefaultSaslConfig, MetricsCollector, SaslConfig}
 import dev.profunktor.fs2rabbit.config.Fs2RabbitConfig
 import dev.profunktor.fs2rabbit.effects.Log
 import dev.profunktor.fs2rabbit.javaConversion._
@@ -33,7 +33,8 @@ object ConnectionResource {
       sslCtx: Option[SSLContext] = None,
       // Unlike SSLContext, SaslConfig is not optional because it is always set
       // by the underlying Java library, even if the user doesn't set it.
-      saslConf: SaslConfig = DefaultSaslConfig.PLAIN
+      saslConf: SaslConfig = DefaultSaslConfig.PLAIN,
+      metricsCollector: Option[MetricsCollector] = None
   ): F[Connection[Resource[F, ?]]] =
     Sync[F].delay {
       new Connection[Resource[F, ?]] {
@@ -47,12 +48,12 @@ object ConnectionResource {
             factory.setVirtualHost(conf.virtualHost)
             factory.setConnectionTimeout(conf.connectionTimeout)
             factory.setAutomaticRecoveryEnabled(conf.automaticRecovery)
-            if (conf.ssl) {
+            if (conf.ssl)
               sslCtx.fold(factory.useSslProtocol())(factory.useSslProtocol)
-            }
             factory.setSaslConfig(saslConf)
             conf.username.foreach(factory.setUsername)
             conf.password.foreach(factory.setPassword)
+            metricsCollector.foreach(factory.setMetricsCollector)
             val addresses = conf.nodes.map(node => new Address(node.host, node.port))
             (factory, addresses)
           }
--- a/site/docs/examples/client-metrics.md
+++ b/site/docs/examples/client-metrics.md
@@ -7,7 +7,7 @@ number: 18
 # Client Metrics
 
 RabbitMQ Java Client supports metrics collection via [Dropwizard](https://www.rabbitmq.com/blog/2016/11/30/metrics-support-in-rabbitmq-java-client-4-0/) or [Micrometer](https://www.rabbitmq.com/blog/2018/04/10/rabbitmq-java-client-metrics-with-micrometer-and-datadog/).
-At the moment of writing both providers are in the `amqp-client` 5.9.0. Just instantiate one as it is shown below.
+At the moment of writing both providers are in the `amqp-client` 5.9.0. You can instantiate one as shown below.
 
 ```scala
 val registry            = new MetricRegistry
--- a/site/docs/examples/client-metrics.md
+++ b/site/docs/examples/client-metrics.md
@@ -1,0 +1,186 @@
+---
+layout: docs
+title:  "Client Metrics"
+number: 18
+---
+
+# Client Metrics
+
+RabbitMQ Java Client supports metrics collection via [Dropwizard](https://www.rabbitmq.com/blog/2016/11/30/metrics-support-in-rabbitmq-java-client-4-0/) or [Micrometer](https://www.rabbitmq.com/blog/2018/04/10/rabbitmq-java-client-metrics-with-micrometer-and-datadog/).
+At the moment of writing both providers are in the `amqp-client` 5.9.0. Just instantiate one as it is shown below.
+
+```scala
+val registry            = new MetricRegistry
+val dropwizardCollector = new StandardMetricsCollector(registry)
+``` 
+
+Now it is ready to use.
+
+```scala
+RabbitClient[IO](config, blocker, metricsCollector = Some(dropwizardCollector))
+```
+
+## Expose via JMX
+
+JMX provides a standard way to access performance metrics of an application. Dropwizard has a module to report metrics
+via JMX with `metrics-jmx` module. Please add it to the list of the dependencies. 
+
+```
+libraryDependencies += "io.dropwizard.metrics" % "metrics-core" % "4.1.5"
+libraryDependencies += "io.dropwizard.metrics" % "metrics-jmx"  % "4.1.5"
+```
+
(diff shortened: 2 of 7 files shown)


kubum:
@gvolpe many thanks for almost instant response! 

<details>
  <summary>I just added `sbt-mina-plugin` to verify master automatically against the latest published 2.1.1</summary>

```git
diff --git a/build.sbt b/build.sbt
index e1565ca..50ccf7d 100644
--- a/build.sbt
+++ b/build.sbt
@@ -103,6 +103,7 @@ lazy val `fs2-rabbit` = project
   .settings(commonSettings: _*)
   .settings(libraryDependencies ++= CoreDependencies(scalaVersion.value))
   .settings(parallelExecution in Test := false)
+  .settings(mimaPreviousArtifacts := Set("dev.profunktor" %% "fs2-rabbit" % "2.1.1"))
   .enablePlugins(AutomateHeaderPlugin)

 lazy val `fs2-rabbit-circe` = project
diff --git a/project/plugins.sbt b/project/plugins.sbt
index c82f0f6..a3fe384 100644
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,10 +1,11 @@
 resolvers += Classpaths.sbtPluginReleases
 resolvers += "Typesafe Repository" at "https://repo.typesafe.com/typesafe/releases/"

-addSbtPlugin("com.geirsson"              % "sbt-ci-release" % "1.5.3")
-addSbtPlugin("de.heikoseeberger"         % "sbt-header"     % "5.6.0")
-addSbtPlugin("com.47deg"                 % "sbt-microsites" % "1.2.1")
-addSbtPlugin("org.scalameta"             % "sbt-mdoc"       % "2.1.4")
-addSbtPlugin("com.scalapenos"            % "sbt-prompt"     % "1.0.2")
-addSbtPlugin("com.lucidchart"            % "sbt-scalafmt"   % "1.16")
-addSbtPlugin("io.github.davidgregory084" % "sbt-tpolecat"   % "0.1.13")
+addSbtPlugin("com.geirsson"              % "sbt-ci-release"  % "1.5.3")
+addSbtPlugin("de.heikoseeberger"         % "sbt-header"      % "5.6.0")
+addSbtPlugin("com.47deg"                 % "sbt-microsites"  % "1.2.1")
+addSbtPlugin("org.scalameta"             % "sbt-mdoc"        % "2.1.4")
+addSbtPlugin("com.scalapenos"            % "sbt-prompt"      % "1.0.2")
+addSbtPlugin("com.lucidchart"            % "sbt-scalafmt"    % "1.16")
+addSbtPlugin("io.github.davidgregory084" % "sbt-tpolecat"    % "0.1.13")
+addSbtPlugin("com.typesafe"              % "sbt-mima-plugin" % "0.7.0")
```
</details>

The output of the `sbt mimaReportBinaryIssues`:

```
[error] (fs2-rabbit / mimaReportBinaryIssues) Failed binary compatibility check against dev.profunktor:fs2-rabbit_2.12:2.1.1! Found 31 potential problems (filtered 7)
```

Amongst all of them, there is the the suspect as we expected:

```
[error]  * method make(dev.profunktor.fs2rabbit.config.Fs2RabbitConfig,scala.Option,com.rabbitmq.client.SaslConfig,cats.effect.Sync,dev.profunktor.fs2rabbit.effects.Log)java.lang.Object in object dev.profunktor.fs2rabbit.algebra.ConnectionResource does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("dev.profunktor.fs2rabbit.algebra.ConnectionResource.make")
```

What do you think? Would it be better to publish 2.2.0 to be safe than sorry? By the way, thank you for caring about those things, I completely forgot about it.

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
```diff
@@ -33,7 +33,8 @@ object ConnectionResource {
       sslCtx: Option[SSLContext] = None,
       // Unlike SSLContext, SaslConfig is not optional because it is always set
       // by the underlying Java library, even if the user doesn't set it.
-      saslConf: SaslConfig = DefaultSaslConfig.PLAIN
+      saslConf: SaslConfig = DefaultSaslConfig.PLAIN,
+      metricsCollector: Option[MetricsCollector] = None
```
Daenyth, line 37: This bincompat change won't affect me personally, since I'm using the higher level constructors iirc

site/docs/examples/client-metrics.md
```diff
@@ -0,0 +1,186 @@
+---
+layout: docs
+title:  "Client Metrics"
+number: 18
+---
+
+# Client Metrics
+
+RabbitMQ Java Client supports metrics collection via [Dropwizard](https://www.rabbitmq.com/blog/2016/11/30/metrics-support-in-rabbitmq-java-client-4-0/) or [Micrometer](https://www.rabbitmq.com/blog/2018/04/10/rabbitmq-java-client-metrics-with-micrometer-and-datadog/).
+At the moment of writing both providers are in the `amqp-client` 5.9.0. Just instantiate one as it is shown below.
```
Daenyth, line 10:
```suggestion
At the moment of writing both providers are in the `amqp-client` 5.9.0. You can instantiate one as shown below.
```

⛏️ Avoid using the word "just" in documentation, as it's always either useless, condescending, or both.

site/docs/examples/client-metrics.md
```diff
@@ -0,0 +1,186 @@
...
+
+```scala
+val registry            = new MetricRegistry
+val dropwizardCollector = new StandardMetricsCollector(registry)
```
Daenyth, line 14: 🤔 Do these need construction in `IO`? In `Resource`? I'd be surprised if they were pure

site/docs/examples/client-metrics.md
```diff
@@ -0,0 +1,186 @@
...
+
+The app is going to have now the following metrics under `com.rabbitmq.client.jmx`:
+* Acknowledged messages
+* Channels count
+* Connections count
+* Consumed messages
+* Published messages
+* Rejected messages
+
+## Full Listing
```
Daenyth, line 76: 🤔 It might be worth us making an examples/ project to hold these kinds of things, but doesn't have to be in this PR

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
kubum: Thank you for confirming! Shall we bump the minor version anyway?

site/docs/examples/client-metrics.md
kubum: Great suggestion, thanks for pointing it out 👍

site/docs/examples/client-metrics.md
kubum: An example Main file was added in the [examples](https://github.com/profunktor/fs2-rabbit/blob/cb3bb15ecb97d8945b14294310d84489d7926b0b/examples/src/main/scala/dev/profunktor/fs2rabbit/examples/DropwizardMetricsDemo.scala). Based on my experience, a lot of people scan the website to build their mental model of what it is first. Do you think it worth having it in the both places?

site/docs/examples/client-metrics.md
kubum: Haha, I didn't find it that bad.

### `MetricRegistry`

https://github.com/dropwizard/metrics/blob/release/4.1.x/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java#L60-L66

### `StandardMetricsCollector`

https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/StandardMetricsCollector.java#L50-L70

Have you experienced any weirdness around this initialisers?

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
gvolpe: @kubum versioning is normally handled separately before a release, don't worry about it :)

site/docs/examples/client-metrics.md
gvolpe: It is usually a good practice to suspend Java stuff because you never know :|

gvolpe:
I'll let @Daenyth do the honors of merging. Thanks for contributing @kubum :)

site/docs/examples/client-metrics.md
Daenyth: 👀 

Those classes contain mutable state; hence construction is not referentially transparent -- you can't factor `new` in/out freely without changing the meaning. It doesn't look like they have a lifecycle though, so no need for `Resource`

site/docs/examples/client-metrics.md
Daenyth: It would be nice if we could embed the file in the doc page, but I don't know if the tool we're using can do that.

It's fine as is I think

kubum: Ah, yeah, I see what you mean. I couldn't find anything like that in the mdoc documentation too.

site/docs/examples/client-metrics.md
kubum: Yeah, they are not resources.

I have no objection to wrap them in IO, I'll do it 👍 

By the way, they are out of IO in the http4s docs too: https://http4s.org/v0.18/middleware/ (scroll to the Dropwizard Metrics Middleware).

Daenyth:
I'll merge now just to get the code in - we can clean up the docs later

PR #407, opened 2020-10-06
Allow to configure requestedHeartbeat

aywengo:
Motivation:

We would like to allow to configure [requestedHeartbeat](https://www.rabbitmq.com/heartbeats.html#using-heartbeats-in-java) parameter for users. 
Default value is 60 seconds. However, some systems are working optimally with values 10-20 sec. 

Modification:
- added optional parameter `requestedHeartbeat` to `Fs2RabbitConfig`
- default value is `60`
- updated doc

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
@@ -47,7 +47,7 @@ object ConnectionResource {
             factory.setPort(firstNode.port)
             factory.setVirtualHost(conf.virtualHost)
             factory.setConnectionTimeout(conf.connectionTimeout)
-            factory.setRequestedHeartbeat(conf.requestedHeartbeat.getOrElse(ConnectionFactory.DEFAULT_HEARTBEAT))
+            factory.setRequestedHeartbeat(conf.requestedHeartbeat)
             factory.setAutomaticRecoveryEnabled(conf.automaticRecovery)
             if (conf.ssl)
               sslCtx.fold(factory.useSslProtocol())(factory.useSslProtocol)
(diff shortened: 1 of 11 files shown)


gvolpe (approved):

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
```diff
@@ -47,6 +47,7 @@ object ConnectionResource {
             factory.setPort(firstNode.port)
             factory.setVirtualHost(conf.virtualHost)
             factory.setConnectionTimeout(conf.connectionTimeout)
+            factory.setRequestedHeartbeat(conf.requestedHeartbeat.getOrElse(60))
```
Daenyth, line 50: ⛏️ I'd like to make it clear where we get the defaults from
```suggestion
            factory.setRequestedHeartbeat(conf.requestedHeartbeat.getOrElse(ConnectionFactory.DEFAULT_HEARTBEAT))
```

core/src/main/scala/dev/profunktor/fs2rabbit/config/Fs2RabbitConfig.scala
```diff
@@ -33,6 +33,7 @@ case class Fs2RabbitConfig(
     requeueOnNack: Boolean,
     requeueOnReject: Boolean,
     internalQueueSize: Option[Int],
+    requestedHeartbeat: Option[Int],
```
Daenyth, line 36: 🤔 What do you think about doing `reqestedHeartbeat: Int = ConnectionFactory.DEFAULT_HEARTBEAT`?

I'm not sure it being optional actually buys us much - maybe it's better to just have it like this.

Daenyth (approved):
I think this is OK to merge, but we'll need to make a new major version - adding fields to a case class isn't backwards compatible.

We should look at maybe using a builder style interface in the future to avoid that (even if we implement it with a case class). Or maybe one of the enhanced case class libraries has something for this

core/src/main/scala/dev/profunktor/fs2rabbit/config/Fs2RabbitConfig.scala
aywengo: Totally agree.

PR #417, opened 2020-10-13
Blocker for all IO operations

wookievx:
All IO-related operations are now invoked via blocker, to avoid leaking worker threads (which might happen if consumer are opened dynamically, several at the same time). This change affect Consume/Ack/Bind/Delete functionality


Daenyth (approved):
Reasonable.

We'll need to be sure to flag it as a breaking release, but it's so easy to fix I'm not too concerned.

PR #421, opened 2020-10-19
ConnectionFactory should be created once for Connection

nikiforo:
I added `println("\n\nI'm creating a new conectionFactory\n\n")` to dev/profunktor/fs2rabbit/algebra/Connection.scala:44,
run the example with `> examples/ test:runMain dev.profunktor.fs2rabbit.examples.FactoryTest`

```
[info] running dev.profunktor.fs2rabbit.examples.FactoryTest 


I'm creating a new conectionFactory


17:28:46.434 [ioapp-compute-0] INFO  d.profunktor.fs2rabbit.effects.Log$ - Acquired connection: amqp://guest@127.0.0.1:5672/


I'm creating a new conectionFactory


17:28:46.444 [ioapp-compute-0] INFO  d.profunktor.fs2rabbit.effects.Log$ - Acquired connection: amqp://guest@127.0.0.1:5672/
```

```
object FactoryTest extends IOApp {

  override def run(args: List[String]): IO[ExitCode] = {
    val config = Fs2RabbitConfig(
      host = "127.0.0.1",
      port = 5672,
      virtualHost = "/",
      username = Some("guest"),
      password = Some("guest"),
      ssl = false,
      connectionTimeout = 3,
      requeueOnNack = false,
      requeueOnReject = false,
      internalQueueSize = Some(500)
    )

    val blockerResource =
      Resource
        .make(IO(Executors.newCachedThreadPool()))(es => IO(es.shutdown()))
        .map(Blocker.liftExecutorService)

    val resource =
      for {
        blocker <- blockerResource
        client  <- Resource.liftF(RabbitClient[IO](config, blocker))
        _       <- client.createConnection
        _       <- client.createConnection
      } yield {}

    resource.use(_ => IO.pure(ExitCode.Success))
  }
}
```

I think that connectionFactory should be created once for ConnectionResource.make, this PR does this.

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
@@ -16,7 +16,6 @@
 
 package dev.profunktor.fs2rabbit.algebra
 
-import cats.data.NonEmptyList
 import cats.effect.{Resource, Sync}
 import cats.implicits._
 import com.rabbitmq.client.{Address, ConnectionFactory, DefaultSaslConfig, MetricsCollector, SaslConfig}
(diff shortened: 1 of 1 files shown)


Daenyth (approved):
A non-bincompat-breaking PR for a change, how refreshing 😁 

:shipit:

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
```diff
@@ -39,41 +38,37 @@ object ConnectionResource {
...
-            if (conf.ssl)
-              sslCtx.fold(factory.useSslProtocol())(factory.useSslProtocol)
-            factory.setSaslConfig(saslConf)
-            conf.username.foreach(factory.setUsername)
-            conf.password.foreach(factory.setPassword)
-            metricsCollector.foreach(factory.setMetricsCollector)
-            val addresses = conf.nodes.map(node => new Address(node.host, node.port))
-            (factory, addresses)
-          }
+        private[fs2rabbit] val connectionFactory = {
```
kubukoz, line 41: I would only suggest to instantiate this outside the `new`, but still in `delay`. But just a suggestion.

nikiforo: This will clarify, why Sync.delay is used.

And it will look like this:
```
Sync.delay{
  val factory = ...
  ...
  factory
}.map { connectionFactory =>
  new Connection ...
}
```
And the functor should be provided somewhere

Have I understood you correctly?

agustafson (approved):
Looks good @nikiforo - thanks!

agustafson:
Any objections if I merge this?

PR #428, opened 2020-11-05
Don't wrap Binding.unbindQueue call in Sync.delay

agustafson:
The call from `unbindQueue` doesn't need wrapping in `Sync.delay` because it's already inside `blocker.delay`

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Binding.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Binding.scala
@@ -68,15 +68,13 @@ object Binding {
                                queueName: QueueName,
                                exchangeName: ExchangeName,
                                routingKey: RoutingKey): F[Unit] =
-        Sync[F].delay {
-          unbindQueue(
-            channel,
-            queueName,
-            exchangeName,
-            routingKey,
-            QueueUnbindArgs(Map.empty)
-          )
-        }.void
+        unbindQueue(
+          channel,
+          queueName,
+          exchangeName,
+          routingKey,
+          QueueUnbindArgs(Map.empty)
+        )
 
       override def unbindQueue(channel: AMQPChannel,
                                queueName: QueueName,


kubukoz:
This is just from a brief look, but it looks to me like it was a nested effect and the one inside was discarded... So this would technically change the behavior. I would need to inspect closer to confirm that though, which i can probably do later today

agustafson:
> This is just from a brief look, but it looks to me like it was a nested effect and the one inside was discarded... So this would technically change the behavior. I would need to inspect closer to confirm that though, which i can probably do later today

Yep, that's right @kubukoz, at the moment it will be returning `F[F[Unit]]`, so the behaviour will change. But I expect that the queue unbinding wouldn't have been occurring given that it's a nested `F[F[_]]`.

I had tried to see if I could write a test to prove that the queue won't be unbound when `unbindQueue` is called, but couldn't find a way of listing all of the currently bound queues (or any other way to prove the unbind happened). I'll take another look to see if I can find a way of testing these changes.

Daenyth:
If you know the name of the queue you should be able to `declare` for it, supplying a value you know contradicts what exists - for example, change the durability or queue type. If the queue exists, then the declaration will fail. If the declaration doesn't fail, you know that the queue didn't exist

agustafson:
> If you know the name of the queue you should be able to `declare` for it, supplying a value you know contradicts what exists - for example, change the durability or queue type. If the queue exists, then the declaration will fail. If the declaration doesn't fail, you know that the queue didn't exist

Thanks @Daenyth, I was wondering about doing the same. I couldn't find anything which could provide any details about which queues existed, so perhaps I'll try updating some tests to check via redeclaration of the queue as you suggest.

agustafson:
I've looked into it and the method I've changed above isn't called by anything.

It seems that there isn't a consistent approach to how the methods are mapped: 
- `RabbitClient` delegates calls for `unbindQueue`, `bindExchange` `unbindExchange` to methods in `RabbitClient` by passing an empty map of arguments
- `Binding` calls `unbindQueue` internally (ie: in `Binding`) by passing an empty map of arguments (but as discussed in the point above, this instance of `unbindQueue` is never called by `RabbitClient`

So we have 2 paths:
- `Binding` needs to be a proxy for the underlying java methods, and so we have a few overloaded methods (such as `bindQueue`, `bindExchange`), one version is passed `Arguments`, and the other version isn't passed `Arguments`. All calls from `RabbitClient` would pass the calls to `Binding` rather than calling overloaded methods in `RabbitClient`
- we minimise the methods in `Binding` so that only the versions with `Arguments` are available, and `RabbitClient` exposes overloaded methods: 1 which passes through `Arguments` provided, and 1 which passes an empty map of `Arguments`

In either case I think we are ok to make these changes as the interface of `RabbitClient` wouldn't change.

jbwheatley:
This overloading in `Binding` of `bindQueue` and `unbindQueue` should just be removed imo. And here https://github.com/profunktor/fs2-rabbit/blob/dd375753af4059c7288508eaa9b93efe7cf26eaf/core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/RabbitClient.scala#L189
should be ` binding.bindQueue(channel, queueName, exchangeName, routingKey, QueueBindingArgs(Map.empty))`. Looks like might have partially been done for `Binding` for the methods `bindExchange` and `unbindExchange`.

agustafson:
> This overloading in `Binding` of `bindQueue` and `unbindQueue` should just be removed imo. Path 2 makes the most sense to me. So here
> 
> https://github.com/profunktor/fs2-rabbit/blob/dd375753af4059c7288508eaa9b93efe7cf26eaf/core/src/main/scala/dev/profunktor/fs2rabbit/interpreter/RabbitClient.scala#L189
> 
> 
> should be ` binding.bindQueue(channel, queueName, exchangeName, routingKey, QueueBindingArgs(Map.empty))`. Looks like might have partially been done for `Binding` for the methods `bindExchange` and `unbindExchange`.

Agreed, I think I like the 2nd option too. Thanks @jbwheatley

PR #434, opened 2020-11-11
Cats effect 3

jbwheatley:


kubukoz (commented):
Looks good, thanks a lot!

jbwheatley:
will hopefully move to M3 at some point this week

jbwheatley:
aha I missed the site docs in my local build!

Daenyth (approved):
Very straightforward!

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Binding.scala
```diff
@@ -68,7 +68,7 @@ object Binding {
                                queueName: QueueName,
                                exchangeName: ExchangeName,
                                routingKey: RoutingKey): F[Unit] =
-        Sync[F].delay {
+        Sync[F].blocking {
```
agustafson, line 71: I think this could just stay the way it was as it wasn't using the `Blocker` previously
```suggestion
        Sync[F].delay {
```

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Consume.scala
```diff
@@ -16,8 +16,8 @@
 
 package dev.profunktor.fs2rabbit.algebra
 
-import cats.effect.syntax.effect._
-import cats.effect.{Blocker, ContextShift, Effect, Sync}
+import cats.effect.kernel.Sync
```
agustafson, line 19: Minor import change
```suggestion
import cats.effect.Sync
```

examples/src/main/scala/dev/profunktor/fs2rabbit/examples/AckerConsumerDemo.scala
```diff
@@ -30,7 +30,7 @@ import dev.profunktor.fs2rabbit.model.ExchangeType.Topic
 import dev.profunktor.fs2rabbit.model._
 import fs2._
 
-class AckerConsumerDemo[F[_]: Concurrent: Timer](fs2Rabbit: RabbitClient[F]) {
+class AckerConsumerDemo[F[_]: Async](fs2Rabbit: RabbitClient[F]) {
```
agustafson, line 33: Think `Flow` below can also change to using `Async` instead of `Concurrent`

agustafson (approved):
Nice one @jbwheatley 👍 

Is it best to make it a WIP until we've got releases of ZIO and Monix to work with? I think it's best to comment out the examples as you've done, and just release something so that people can start experimenting with it.

kubukoz:
We could have a new branch for this. But first we should discuss whether we want to maintain 2.x-compatible versions once CE 3 is released.

@gvolpe @agustafson do you know any production users of this library? It would be good to ask them about their migration plans so we can possibly take that into consideration.

Daenyth:
I use it in prod. For me, I think it would be fine to just maintain the one. And honestly, if we cared to, we could backport bugfixes if they're urgent.

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Binding.scala
jbwheatley: this is actually the place where you noticed that this delay is discarding the unbind #428 , so needs to be removed

jbwheatley:
Also use it in prod, and second @Daenyth's comment

examples/src/main/scala/dev/profunktor/fs2rabbit/examples/AckerConsumerDemo.scala
jbwheatley: `Async <: Concurrent` in ce3

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Binding.scala
agustafson: Yeah, I think in this case it prolly still makes sense to keep it as it was before, and let those changes be made separately.

agustafson:
Jack and I both work at the same place and we use fs2-rabbit a fair bit.

Agreed, think we're best to just have a backport branch for the previous version. I think we'll probably need to support it for a while.

kubukoz:
Hey @jbwheatley, there are some compilation issues, it's not anything that should be a problem in the migration, right?

jbwheatley:
> Hey @jbwheatley, there are some compilation issues, it's not anything that should be a problem in the migration, right?

@kubukoz Yeah, just waiting for some stuff from the next fs2 milestone, off the top of my head. Also some examples won't compile until things like monix get a  cats-effect interop for ce3

kubukoz:
oh, that'll take a while since apparently they didn't start yet - I suppose we might need to skip the monix tests for a while.

PR #454, opened 2020-12-18
Use `ConsumerArgs` where possible instead of passing 4 parameters

agustafson:
It looks like we're passing 4 parameters around in multiple places, but we could simply use `ConsumerArgs`.

- Use `ConsumerArgs` instead of passing 4 parameters
- Move default method parameters to `ConsumerArgs.default`

This is a non-breaking change as I've kept the existing methods, marked them as deprecated, and they call the new methods I've defined.


agustafson:
Have checked this with the MiMa tool and it failed, so will fix those errors or mark them as unimportant

agustafson:
@mergifyio rebase

PR #501, opened 2021-03-03
Reducing `RabbitClient`s dependencies

jbwheatley:
Some consolidation of the classes wrapped by `RabbitClient`

- some algebras creations were unnecessarily wrapped in `Sync[F].delay`
- RabbitClient required a `Publish` and a `Publishing`, but `PublishingProgram` extends both of those
- RabbitClient required a `Consuming` and a `Consume`, and `AckConsumingProgram` filled the requirement of both except for one method, `basicCancel`, so I pulled that into its own trait `Cancel`, and had `AckConsumingProgram` extend `Cancel`, so it can replace the two consume dependencies

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Cancel.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Cancel.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2017-2020 ProfunKtor
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Cancel.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.profunktor.fs2rabbit.algebra
+
+import dev.profunktor.fs2rabbit.model.{AMQPChannel, ConsumerTag}
+
+trait Cancel[F[_]] {
```
Daenyth, line 21: ⛏️ I think it would be helpful to add a little class/method doc here

Daenyth (approved):
Seems reasonable

PR #502, opened 2021-03-04
rabbit client ops - imapK, liftAttemptK

jbwheatley:
A means of lifting a rabbit client into a different effect context, and an implementation of lifting from `F[_]` to `EitherT[F, Throwable, *]`. 

New features are in `../interpreter/package.scala`, with `imapK` and `mapK` methods defined for each of `RabbitClient`'s dependencies to facilitate these.


agustafson (approved):
LGTM 👍

Issue #461, opened 2020-12-29
Allow "multiple" ack flag to be reset per call


Daenyth:
Right now, the "ack multiple messages" property gets set on creation, but I believe the api allows it to be set per call. That would be better, because it allows an application to use mixed modes where convenient. I think it's reasonable to keep the creation-time setting as a default, and have overrides that allow it to be set later.

jbwheatley:
fixed in 4.0.0 (requires cats-effect 3)

PR #503, opened 2021-03-04
create acker consumer with configurable ack-multiple flag

jbwheatley:
Addresses #461

- adds a new function to rabbit client: `createAckerConsumerWithMultipleFlag` which returns an acker function `(AckResult, AckMultiple) => F[Unit]` args as oppose to `createAckerConsumer`'s return type of `AckResult => F[Unit]`. 

- adds an arg to `createAckerConsumer` to set the ack multiple flag for the entire lifetime of the created consumer. 

(sorry, scalafmt rules are on a rampage since I turned it on in my ide)

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Acking.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Acking.scala
@@ -16,8 +16,9 @@
 
 package dev.profunktor.fs2rabbit.algebra
 
-import dev.profunktor.fs2rabbit.model.{AMQPChannel, AckResult}
+import dev.profunktor.fs2rabbit.model.{AMQPChannel, AckMultiple, AckResult}
 
 trait Acking[F[_]] {
-  def createAcker(channel: AMQPChannel): F[AckResult => F[Unit]]
+  def createAcker(channel: AMQPChannel, ackMultiple: AckMultiple): F[AckResult => F[Unit]]
+  def createAckerWithMultipleFlag(channel: AMQPChannel): F[(AckResult, AckMultiple) => F[Unit]]
 }
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/program/AckingProgram.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/program/AckingProgram.scala
@@ -36,13 +36,22 @@ case class WrapperAckingProgram[F[_]: Sync] private (
     config: Fs2RabbitConfig,
     consume: Consume[F]
 ) extends AckingProgram[F] {
-  override def createAcker(channel: AMQPChannel): F[AckResult => F[Unit]] = Applicative[F].pure {
-    case Ack(tag) => consume.basicAck(channel, tag, multiple = false)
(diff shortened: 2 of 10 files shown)


core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Acking.scala
```diff
@@ -16,8 +16,9 @@
 
 package dev.profunktor.fs2rabbit.algebra
 
-import dev.profunktor.fs2rabbit.model.{AMQPChannel, AckResult}
+import dev.profunktor.fs2rabbit.model.{AMQPChannel, AckMultiple, AckResult}
 
 trait Acking[F[_]] {
-  def createAcker(channel: AMQPChannel): F[AckResult => F[Unit]]
```
agustafson, line 22: Might need to mark this as deprecated first, or are we happy making it a breaking change?

We could keep the existing method as
```scala
def createAcker(channel: AMQPChannel): F[AckResult => F[Unit]] =
   createAcker(channel: AMQPChannel, AckMultiple(false))
```
(if we allow an overloaded name)

core/src/main/scala/dev/profunktor/fs2rabbit/program/AckingProgram.scala
```diff
@@ -36,14 +36,23 @@ case class WrapperAckingProgram[F[_]: Sync] private (
...
 
+  override def createAckerWithMultipleFlag(channel: AMQPChannel): F[(AckResult, AckMultiple) => F[Unit]] =
+    Applicative[F].pure {
+      case (Ack(tag), flag) => consume.basicAck(channel, tag, multiple = flag.multiple)
+      case (NAck(tag), flag) =>
+        consume.basicNack(channel, tag, multiple = flag.multiple, config.requeueOnNack)
+      case (Reject(tag), _) =>
+        consume.basicReject(channel, tag, config.requeueOnReject)
+    }
+
```
agustafson, line 55: Unfortunately can't apply this suggestion to the whole code block, but figured it was safer to have the duplicate logic in a method
```suggestion
  private def ackResult(channel: AMQPChannel, multiple: AckMultiple): AckResult => F[Unit] = {
    case Ack(tag) => 
      consume.basicAck(channel, tag, multiple = multiple.multiple)
    case NAck(tag) =>
      consume.basicNack(channel, tag, multiple = multiple.multiple, config.requeueOnNack)
    case Reject(tag) =>
      consume.basicReject(channel, tag, config.requeueOnReject)
  }

  override def createAcker(channel: AMQPChannel, multiple: AckMultiple): F[AckResult => F[Unit]] =
    Applicative[F].pure {
      ackResult(channel, multiple)
    }

  override def createAckerWithMultipleFlag(channel: AMQPChannel): F[(AckResult, AckMultiple) => F[Unit]] =
    Applicative[F].pure {
      case (result, flag) => ackResult(channel, flag).apply(result)
    }
```

agustafson (approved):
LGTM 👍

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Acking.scala
jbwheatley: I think we can make it a breaking change, as this will go into the next major release. The API exposed by `RabbitClient` should remain source-compatible, so I'm not too worried.

PR #702, opened 2022-03-03
feature: handle client-provided connection name

ldip:
integrate client-provided connection name into Fs2RabbitConfig
	give client-provided connection name inside acquireConnection


core/src/main/scala/dev/profunktor/fs2rabbit/config/Fs2RabbitConfig.scala
```diff
@@ -39,7 +39,8 @@ case class Fs2RabbitConfig(
     requeueOnReject: Boolean,
     internalQueueSize: Option[Int],
     requestedHeartbeat: FiniteDuration,
-    automaticRecovery: Boolean
+    automaticRecovery: Boolean,
+    clientProvidedName: Option[String]
```
jbwheatley, line 43: `clientProvidedConnectionName` instead maybe? That way its easier for users to find the relevant bit of rabbit docs if they don't know what this field does.

ldip: Yes I do this, nice idea !

PR #819, opened 2023-03-10
Add support for basic.get

ksonj:
This PR adds support for [AMQPs `basic.get`](https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.get) method. I tried to isolate the style changes that were applied automatically in a separate commit. If you prefer separate PRs for those, please just say so.

This moves some methods that were previously defined in `Consume` into a new trait `Get` (along with the newly added `get`), which `Consume` now extends. I have to admit I was not completely sure as to what the intended separation of concerns is with all the `<Verb>`, `<Verb>ing` and `<Verb>Program` traits. I hope I didn't mess things up too very much.


ksonj:
Seems I broke binary compatibility in a few places. If there's interest to merge, I'd be happy to see what I can do about that.

PR #908, opened 2024-01-08
Blocking state fix + update copyright year

matejcerny:
Fix for `[WARNING] A Cats Effect worker thread was detected to be in a blocked state (TIMED_WAITING)`


bilejpes (approved):
Thanks, was waiting for the fix for a while.

PR #953, opened 2024-05-22
Dynamic versioning via dynver

matejcerny:


geirolz:
Seems JFrog is having maintenance - Idk if it is related to the failure on downloading sbt 
https://status.jfrog.io/

PR #955, opened 2024-05-26
Refactoring build.sbt

geirolz:
@matejcerny I tried to replicated what I have I my libraries, Idk if it works since I miss the approval to run the pipeline

UPDATE
Just realized you already added `crossScalaVersions := Nil`, I'll still have a look in the mean while if could I have the rights to run the pipeline without the approval it would be less painful for both I guess

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import Dependencies._
 import microsites.ExtraMdFileConfig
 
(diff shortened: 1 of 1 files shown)


matejcerny:
I added you the rights, feel free to create/delete the "testing" tag `v5.1.2`

PR #956, opened 2024-05-28
Add headers type

geirolz:
This PR add the type `Headers` and the types aliases `HeaderKey = String` and   `HeaderKey = (String, AmqpFieldValue)`

It also adds `AmqpFieldDecoder` and `AmqpFieldEncoder` to decode and encode `AmqpFieldValue` values into scala models

The operator `:=` is provided as sugar syntax as extension method for `HeaderKey` ( String ), 
useful instead of using `->` because creates a tuple of `(String, AmqpFieldValue)` encoding a generic type `T`. 
I'm not sure to profile this method for strings since could potentially collide with other libraries or user definitions, maybe it would be worth to define `HeaderKey` as a value class providing a string context extension method like `val key: HeaderKey = hk"MY_HEADER_KEY"`. What do you thing ?

As final step of this PR i've simply extracted some classes from the `model` object that has grow and became too big to handle. 

Here some example:

```scala
import dev.profunktor.fs2rabbit.model.Headers
import dev.profunktor.fs2rabbit.model.Headers._
 
val headers = Headers(
      "x-custom-header"   := "value", // StringVal
      "x-custom-header-2" := 123 // IntVal
)
.append("x-custom-header-3", 123L) // LongVal
.remove("x-custom-header-3")

/**
* OR
*   val headers = Headers(
*    "x-custom-header" -> StringVal("value"),
*    "x-custom-header-2" -> IntVal(123)
*   )
*   .append("x-custom-header-3", LongVal(123L))
*   .remove("x-custom-header-3")
* */

headers.getRaw[IO]("x-custom-header") // IO(StringVal("value"))
headers.getOptRaw("x-custom-header") // Some(StringVal("value"))

headers.getAs[IO, String]("x-custom-header") // IO("value")
headers.getOptAs[String]("x-custom-header") // Some("value")
headers.getOptAsF[IO, Int]("x-custom-header-2") // IO(Some(123))
```


geirolz:
> Quite big, not sure if I have time soon for it 😞

Np, let's release now and have a look to this when you have time. 
I've changed a lot but I introduced just one braking change, `AmqpProperties` now wants an instance of `Headers` instead of `Map`. I'll try to summarize all changes in the description in order to simplify the review

geirolz:
@matejcerny I've updated the description. I hope it helps

matejcerny (approved):

PR #972, opened 2024-09-03
Added automaticTopologyRecovery to Fs2RabbitConfig

SheliakLyr:
TopologyRecovery feature in amqp-client is sometimes strange. I prefer to have it disabled. This PR adds the needed configuration option to Fs2RabbitConfig.

This change is not backwards compatible in Fs2RabbitConfig constructor. It is not possible to add a default - only one apply method can have defaults defined. I could try to add some additional methods to keep compability.

Also, I fixed a small compilation error (unrelated).


core/src/main/scala/dev/profunktor/fs2rabbit/config/Fs2RabbitConfig.scala
```diff
@@ -57,6 +58,7 @@ object Fs2RabbitConfig {
       internalQueueSize: Option[Int],
       requestedHeartbeat: FiniteDuration = FiniteDuration(ConnectionFactory.DEFAULT_HEARTBEAT, TimeUnit.SECONDS),
       automaticRecovery: Boolean = true,
+      automaticTopologyRecovery: Boolean = true,
```
geirolz, line 61: This is the default value for the java library, I suppose, is it not?

geirolz:
> This change is not backwards compatible in Fs2RabbitConfig constructor

This isn't a problem I'd say

core/src/main/scala/dev/profunktor/fs2rabbit/config/Fs2RabbitConfig.scala
SheliakLyr: Yes, in "modern" versions of that library.

geirolz: Yea, I mean - the one we are using in fs2-rabbit at least

geirolz:
P.S. Ignore the CI error, there is a PR to fix it and it isn't related to your changes

geirolz (approved):

matejcerny (approved):

PR #978, opened 2024-09-06
Cross compile to scala 3.5.x

geirolz:


matejcerny (approved):
Are there any advantages? Just curious...

geirolz:
> Are there any advantages? Just curious...

Just checking compatibility with 3.5.x and potentially use 3.5.x features ( in a dedicated package ) 
Moving towards to scala 3 in general

matejcerny:
> > Are there any advantages? Just curious...
> 
> Just checking compatibility with 3.5.x and potentially use 3.5.x features ( in a dedicated package ) Moving towards to scala 3 in general

What will be the next step in general? Will we keep the 3.5.2 or replace it with the 3.6.2?

geirolz:
I would cross compile all major and minor since it is almost cost free for this kind of library. 
Moreover I'd drop 2.12 in order to just support 2.13 and all 3.x from now on. 

Ideally we have to drop the oldest version when is no more commonly used, this also depends on how easy it is to migrate 3.x minors I'd say ( for the community I mean ). 

I think we cannot define a strict rule now, let's see how Scala evolves and how difficult It becomes to maintain this library supporting multiple versions.

PR #988, opened 2024-09-19
Add methods to setup settings fluently

geirolz:
Adds some methods, just aliases to copy, to setup settings fluently.

I have two doubts: 
- Is it better to use `with` or `as` ? 
- Is it better to accept booleans instead of `Durable/NonDurable`, etc.. ? Like `.withDurable(true)`


matejcerny (approved):
My personal preference is to not use booleans.

Issue #990, opened 2024-09-24
Feature Request: Add way to setMaxInboundMessageBodySize


tboon-cfgms:
As data payloads get larger and larger, the default Java client value for maxInboundMessageSize of 64MB is too limiting for some use cases. AMQP spec defines the max message size as 512MB (see https://www.rabbitmq.com/docs/configure#config-items `max_message_size`). The Java client exposes `setMaxInboundMessageSize` as a method on `ConnectionFactory` - `Fs2RabbitConfig` should allow this value to be set and default to the Java client default of 64MB if not provided.

As an extension of this, add more optional configuration to allow any value in `ConnectionFactory` to be set beyond those currently supported. But that is a bigger ask.

I would be willing to make a PR for `maxInboundMessageSize`.

geirolz:
Hi @tboon-cfgms thanks for reporting this! 
I think we can have both quite easily but let's start with just `maxInboundMessageSize` first I would say

A PR is welcome for sure

PR #991, opened 2024-09-26
Custom conn factory and support maxInboundMessageBodySizeBytes

geirolz:
Address #990

### DIFF
--- a/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
+++ b/core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
@@ -125,7 +125,7 @@ object ConnectionResource {
   ): F[Connection[Resource[F, *]]] =
     Sync[F]
       .delay {
-        val factory   = new ConnectionFactory()
+        val factory   = conf.connectionFactory.getOrElse(new ConnectionFactory())
         val firstNode = conf.nodes.head
         factory.setHost(firstNode.host)
         factory.setPort(firstNode.port)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
```diff
@@ -125,7 +125,7 @@ object ConnectionResource {
   ): F[Connection[Resource[F, *]]] =
     Sync[F]
       .delay {
-        val factory   = new ConnectionFactory()
+        val factory   = conf.connectionFactory.getOrElse(new ConnectionFactory())
```
matejcerny, line 128: I would rather have one make function with a connection factory and another with configs

core/src/main/scala/dev/profunktor/fs2rabbit/config/Fs2RabbitConfig.scala
```diff
@@ -40,10 +40,15 @@ case class Fs2RabbitConfig(
     requestedHeartbeat: FiniteDuration,
     automaticRecovery: Boolean,
     automaticTopologyRecovery: Boolean,
-    clientProvidedConnectionName: Option[String]
+    clientProvidedConnectionName: Option[String],
+    maxInboundMessageBodySizeBytes: Int,
+    connectionFactory: Option[ConnectionFactory]
```
matejcerny, line 45: If there will be two make functions, it's not needed to mix config values with a java connection factory

core/src/main/scala/dev/profunktor/fs2rabbit/algebra/Connection.scala
geirolz: It isn't that easy unfortunately - I'll try again asap

Issue #1004, opened 2024-10-28
CI/CD: Add a scalafmt check to github actions


matejcerny:
To prevent errors with Sonatype release, add a `scalafmtAll` step to github actions.

geirolz:
Do we want a step that fails if the code isn't properly formatted ? Because `scalafmtAll` implies a commit

matejcerny:
Sorry, you are right, it should be `scalafmtCheckAll`. Imho it's better to fail the whole pipeline rather than examine why the release does not work.

PR #1007, opened 2024-10-29
Add scalafmt workflow

geirolz:
This PR address Issue #1004 creating a workflow for every PR. Then we could consider making this check mandatory to merge.


matejcerny (approved):
Good job!

